PR #2, opened 2023-02-10
Bytecode backend

PalumboN:
This PR includes:

- New Pharo Bytecode backend.
- Improvements on Inline optimisation, in order to support other optimisations on CFG
- Fix some bugs
- Fix blocks access and some cfg assertions in tests
- More...

## New return model

I changed how returned values are attached to the frames in order to support non-locals returns after deferred inlines (and optimisations run in between).

- New instruction: `DRStrongCopy` is a copy but immune to optimisations. Converts into a copy when there are no more message sends in the cfg.
- New instruction: `DRFrameReturn` is a strong copy that knows the method that returns. It is used to have information about the returned values in the cfg.

A `^` is interpreted by IRGenerators as follows:

1. **Frames are created** with a default value (receiver or self) and a `FrameMirror`. All frames belonging to the same node share the same mirror object.
1. **Return node** adds deferred returns to the method frame. These deferred returns are saved in the mirror. A **deferred return** is a _copy of the current state (with the returned value pushed in the operand stack)_ and its corresponding _frame return instruction_.
1. When a **Frame is popped**, all deferred returns (shared by all frames) are solved (merged in a phi) and saved in the mirror. It also cleans all deferred returns (they were already included in the cfg).
1. When a **message send** is created, the IRGenerator attaches a copy of its state (including copies of the frames), to be used for future inlining.
1. When a **Frame is popped**, it also checks if a _popped value_ already exists in the mirror. If exists then the frame was popped before and the algorithm merges current deferred returns in a phi function and then merges that phi with the previous popped value.


Druid/DRRegisterCollapse.class.st
```diff
@@ -0,0 +1,13 @@
...
+	#category : #'Druid-BytecodeGeneration'
+}
+
+{ #category : #accessing }
+DRRegisterCollapse >> applyTo: cfg [
+
+	cfg instructionsDo: [ :i | 
+		i isMandatoryInstruction ifFalse: [ 
+			i basicBlock removeInstruction: i ] ]
+]
```
PalumboN, line 13: This is here because the transformation from register-based model (Druid) to a stack-based (Bytecode).

Now, many instructions are removed from CFG to be iterated in different way.

We need to look a ~better~ real strategy for this. Check https://www.sciencedirect.com/science/article/pii/S1571066111001538?ref=pdf_download&fr=RR-2&rr=7598ca93ec5e7761

Druid/DRStackFrame.class.st
```diff
@@ -219,16 +226,60 @@ DRStackFrame >> receiverWithState: aState [
...
+
 	^ returnValue
 ]
 
 { #category : #accessing }
 DRStackFrame >> returnValue: aValue [
-	
-	returnValue := aValue
+
+	"aValue isFrameReturn ifTrue: [ 1halt ]."
```
PalumboN, line 280: Commented code

Druid/DRStackFrame.class.st
```diff
@@ -219,16 +226,60 @@ DRStackFrame >> receiverWithState: aState [
...
+			mirrorFrame returnValue.
+			deferredFrameReturn }.
+	mirrorFrame returnValue: (newFrameReturn).
+	self clearDeferredMethodReturns
+]
+
 { #category : #accessing }
 DRStackFrame >> returnValue [
 
+	"	^ mirrorFrame returnValue"
```
PalumboN, line 272: Commented code

PalumboN (commented):
Review w/ EL @guillep

guillep (approved):

PR #3, opened 2023-02-14
Task108

hernanmd:
Add an object to hold primitive compilation metadata: 
    Primitive number
    Selector of interpreter method
    Selector of generated method (gen...)
    Number of arguments
    CFG

Handle integers in primtive table, for example including -1:

Create a class which receives an interpreter class (for now called DRInterpreterCompiler) with primitive methods, compiles them, and generate a model (before dumping to code). We use a model to avoid managing/cleaning code generation artifacts. From there we could dump the code in a later stage. This new class is tested in DRInterpreterToCompilerTest.


Druid/CCodeGenerator.extension.st
```diff
@@ -0,0 +1,16 @@
+Extension { #name : #CCodeGenerator }
+
+{ #category : #'*Druid' }
+CCodeGenerator >> inferTypes [
+
```
guillep, line 5: This method should not be moved to this package!

Druid/CompiledMethod.extension.st
```diff
@@ -0,0 +1,13 @@
+Extension { #name : #CompiledMethod }
+
+{ #category : #'*Druid' }
+CompiledMethod >> numberOfArguments [
+
```
guillep, line 5: Be careful, this method name could override  a method in Pharo and break everything...

Druid/DRAbstractMethodCompilerCompiler.class.st
```diff
@@ -30,37 +30,38 @@ DRCompilerCompiler >> compile [
...
 
 	| startBlock initialBasicBlock |
+
 	"Create the initial basic block"
 	startBlock := generator initialBasicBlock.
 	initialBasicBlock := generator createInitialBasicBlock.
 	startBlock jumpTo: initialBasicBlock.
 
-	generator numberOfArguments: (self numberOfArgumentsFor: method)
+	generator numberOfArguments: method numberOfArguments
```
guillep, line 54: This is why the tests are breaking and there is the problem with the numberOfArguments pragma. This method should be defined on the compiler, because different compilers assume different ways to define number of arguments.

I believe this line should be reverted

Druid/DRBasicCogInterpreterObjectMemoryPrimitives.class.st
```diff
@@ -0,0 +1,25 @@
+Class {
+	#name : #DRBasicCogInterpreterObjectMemoryPrimitives,
+	#superclass : #CogVMSimulatorLSB,
```
guillep, line 3: Is this class used?

Druid/DRBasicCogInterpreterPrimitivesWithoutParameters.class.st
```diff
@@ -0,0 +1,26 @@
+Class {
+	#name : #DRBasicCogInterpreterPrimitivesWithoutParameters,
+	#superclass : #CogVMSimulatorLSB,
```
guillep, line 3: Same here

Druid/DRCompilationBuildUnit.class.st
```diff
@@ -0,0 +1,39 @@
+Class {
+	#name : #DRCompilationBuildUnit,
```
guillep, line 2: Maybe should this be called compilation unit builder?

Druid/DRPrimitiveCompilerCompiler.class.st
```diff
@@ -50,9 +67,11 @@ DRPrimitiveCompilerCompiler >> newIRGenerator [
...
 ]
 
-{ #category : #accessing }
-DRPrimitiveCompilerCompiler >> numberOfArgumentsFor: aMethod [ 
-	^ (aMethod ast pragmaNamed: #numberOfArguments:) arguments first value
+{ #category : #'generation-IR' }
+DRPrimitiveCompilerCompiler >> numberOfArgumentsFor: anObject [
+	"Answer a <Number> representing the number of arguments found in anObject (method or selector in the receiver's primitive's source)"
+
+	^ anObject numberOfArgumentsIn: self
```
guillep, line 74: I thing this method (numberofargumentsin:) should be renamed or inlined as it was before. The fact that it takes the pragma is important for this implementation class, and it should not be important for other compilers

Druid/Exception.extension.st
```diff
@@ -0,0 +1,9 @@
+Extension { #name : #Exception }
+
+{ #category : #'*Druid' }
+Exception >> handlePrimitiveCompilationFor: aDRPrimitiveCCCollector [
+
+	aDRPrimitiveCCCollector
+		failedPrimitivesAt: self
+		add: aDRPrimitiveCCCollector sourceName
```
guillep, line 8: Is the extension method really justified here? There is no polymorphism, nor access to the exception state, right ?

Druid/PCGArrayNode.extension.st
```diff
@@ -0,0 +1,7 @@
+Extension { #name : #PCGArrayNode }
+
+{ #category : #'*Druid' }
+PCGArrayNode >> maxCompiledPrimitives [
```
guillep, line 4: Why on array ?

Druid/Symbol.extension.st
```diff
@@ -1,5 +1,11 @@
 Extension { #name : #Symbol }
 
+{ #category : #'*Druid' }
+Symbol >> numberOfArgumentsIn: aDRPrimitiveCompilerCompiler [
+
+	^ aDRPrimitiveCompilerCompiler sourceMethod numberOfArguments
```
guillep, line 6: Ok, so there is polymorphism, but I don't see how this work for primitive compilation, because there is no access to the pragma here!

Druid/VMClass.extension.st
```diff
@@ -0,0 +1,8 @@
+Extension { #name : #VMClass }
+
+{ #category : #'*Druid' }
+VMClass class >> resetInitializationOptions [
+
+	super release.
+	InitializationOptions := nil
```
guillep, line 7: If this is important for the VM, let's push it to the vm :)

Druid/VMMaker.extension.st
```diff
@@ -0,0 +1,8 @@
+Extension { #name : #VMMaker }
+
+{ #category : #'*Druid' }
+VMMaker >> stopOnErrors [
+
```
guillep, line 5: Same here

guillep (changes requested):
I think I spotted why the tests break.

I like the comments!!

I see an abuse of extension methods, we should set ground rules for them I think :)

Druid/CompiledMethod.extension.st
hernanmd: I checked in a Pharo image with Druid from main branch:

```smalltalk
#numberOfArguments implementors "an OrderedCollection(DRBlockClosure>>#numberOfArguments DRCogitSendMarshall>>#numberOfArguments DRJITMessageSend>>#numberOfArguments TraitMethodDescription>>#numberOfArguments VMCompiledCodeBuilder>>#numberOfArguments)"
```

so it seems isn't breaking anything but for method extensions in general I would prefix with #dr ...all methods since these kind of selectors could break core tools. What do you think?

Druid/DRBasicCogInterpreterObjectMemoryPrimitives.class.st
hernanmd: There was an autocompletion mistake I did. Now it's referenced in #interpreterClassObjectMemoryPrimitives

Druid/DRBasicCogInterpreterPrimitivesWithoutParameters.class.st
hernanmd: It is referenced now in #interpreterClassUnaryPrimitives

Druid/Exception.extension.st
hernanmd: I don't understand why there should be polymorphism or access to the exception state?

Druid/VMMaker.extension.st
hernanmd: I don't know about this

#stopOnErrors in PharoVMMaker and SlangTyper for some reason don't provide a default?

I will check again deleting these methods, and evaluate with the new compilation we did today

Druid/PCGArrayNode.extension.st
hernanmd: It answers the number of "definitions" (elements) that specify a primitive in the Array of primitives table. This is the parameter of the "from:" in #initializePrimitiveTable

Druid/DRAbstractMethodCompilerCompiler.class.st
hernanmd: I don't like this (all implementors of #numberOfArgumentsFor: just use the method passed as parameter to send a message, no single use of the compiler for this...) but I'll revert it anyway.

guillep: I understand! But in my image I have three implementors of it, and with different implementations.

imagen

We can improve it, but I'm saying that probably this change is the cause of the broken tests in this PR.

guillep: And they are sending a message to the method, yes, but doing different things nevertheless :)

Druid/CCodeGenerator.extension.st
hernanmd: Fixed

Druid/DRInterpreterBuilder.class.st
```diff
@@ -0,0 +1,81 @@
...
+DRInterpreterBuilder >> interpreterClassObjectMemoryPrimitives [
+
+	^ DRBasicCogInterpreterObjectMemoryPrimitives
+]
+
+{ #category : #'accessing - interpreters' }
+DRInterpreterBuilder >> interpreterClassUnaryPrimitives [
+
+	^ DRBasicCogInterpreterPrimitivesWithoutParameters
+]
```
PalumboN, line 55: Why three different classes?

Druid/DRInterpreterCompiler.class.st
```diff
@@ -0,0 +1,310 @@
...
+- All the generated bytecodes or primitives are unimplemented.
+- No error handling.
+
+This enables to build a VM which includes the JIT, but does not know how to JIT any method. Then the primitives or bytecodes could be added incrementally, and provides a baseline to compare performance between implementations.
+
+
+
+"
+Class {
+	#name : #DRInterpreterCompiler,
```
PalumboN, line 13: This is not a Compiler. This uses the Compiler(Compiler). 
It's a Compilation or CompilationBuilder.

Druid/DRPrimitiveCompilerCompilerCollector.class.st
```diff
@@ -0,0 +1,41 @@
+"
+This is a helper class to collect primitive method which couldn't be compiled by Druid.
+
+"
+Class {
+	#name : #DRPrimitiveCompilerCompilerCollector,
+	#superclass : #DRPrimitiveCompilerCompiler,
```
PalumboN, line 7: A collector shouldn't be another object? Instead of subclass?

Maybe this should be move to the any compilation builder or pushed to the superclass

Druid/DRPrimitiveIRGenerator.class.st
```diff
@@ -1,3 +1,8 @@
+"
+Expects a primitive method, this is, a method of an interpreter implementing a primitive (e.g. #primitiveAdd) and answer an IR, useful to generate code.
```
PalumboN, line 2:
```suggestion
Expects a primitive method, this is a method of an interpreter implementing a primitive (e.g. #primitiveAdd), and generates a Druid's IR.
```

PalumboN: I changed "answer" to "generates" because this produces an effect (and I like to use "answer" queries)

Druid/DRPrimitiveObject.class.st
```diff
@@ -0,0 +1,108 @@
...
+	#instVars : [
+		'sourceSelector',
+		'genSelector',
+		'argumentCount',
+		'cfg',
+		'primitiveNumber',
+		'mayCallback'
+	],
+	#category : #'Druid-InterpreterBuilder'
+}
```
PalumboN, line 21: Who is using this class?

Druid/DRCogitPrimitiveInstaller.class.st
```diff
@@ -0,0 +1,154 @@
...
+"
+Class {
+	#name : #DRCogitPrimitiveInstaller,
+	#superclass : #Object,
+	#instVars : [
+		'primitiveSpec',
+		'primTableNode'
+	],
+	#category : #'Druid-InterpreterBuilder'
+}
```
PalumboN, line 15: Is it tested? That the generated method is ok.

Druid/Exception.extension.st
PalumboN: Because if not why is this method here :P

Druid/OrderedCollection.extension.st
```diff
@@ -1,5 +1,10 @@
 Extension { #name : #OrderedCollection }
 
+{ #category : #'*Druid' }
+OrderedCollection >> asPCG [
+	^ PCGArrayNode statements: (self collect: #asPCG)
+]
```
PalumboN, line 6: image

Druid/PCGArrayNode.extension.st
PalumboN: Mmmmmm I think that we have different ways to select the responsibilities of objects. 🤔 

I don't understand why we should ask this to the node and not to any Compilation object who already tracks the primitives to compile.

PalumboN (commented):
I left more comments. There are many extension methods with logic attached to another classes where they are being declared (kind of _feature envy_).

Druid/Exception.extension.st
hernanmd: This technique allows different exceptions to respond to primitive compilation errors in different ways. A more generic collector (like this one) can simply gather all the exceptions in a Dictionary, so that they can be inspected later without having to suffer the constant opening of a Debugger on each compilation error. Due to the stage of the project this happens quite often if one wants to implement a tool for a minimal inspection of the implementation (outside the tests which are a safe environment and where one is sure that everything will work as expected). This logic is implemented in several places in the image (e.g. #manageTestProcessBy:, #notifyUserOfCommand:, #recordResultOf: inHDTestReport:, #stDebuggerInspectorNodesFor:). It also allows it to be re-implemented more generically for any type of "AbstractMethodCompilerCompiler", e.g. DRMethodCompilerCompiler, DRBytecodeCompilerCompiler.

One thing I can do is to move this class to a ToolsSupport package and thus not "mess up" the current implementation.

Druid-Tests/DRCogitPrimitiveInstallerTest.class.st
```diff
@@ -0,0 +1,36 @@
...
+		'primitiveMethod'
+	],
+	#category : #'Druid-Tests-CompilerCompiler'
+}
+
+{ #category : #running }
+DRCogitPrimitiveInstallerTest >> buildUnitForTesting [
+
+	^ DRCompilationBuildUnit new
+]
```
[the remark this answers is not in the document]
hernanmd: Yes, I will remove it for the next commit

Druid/CCodeGenerator.extension.st
```diff
@@ -0,0 +1,8 @@
+Extension { #name : #CCodeGenerator }
+
+{ #category : #'*Druid' }
+CCodeGenerator >> stopOnErrors [
+
+	^ stopOnErrors
+		ifNil: [ stopOnErrors := true ]
+]
```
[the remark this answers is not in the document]
hernanmd: Why not?

Druid/DRPrimitiveObject.class.st
hernanmd: See ```st
DRInterpreterCompiler>>newPrimitive:
```

File: Druid-Tests/DRInterpreterToCompilerTest.class.st [hunk 22]
```diff
@@ -0,0 +1,89 @@
...
+	self assert: interpreterCompiler primitives size equals: interpreterBuilder primitivesCount
+]
+
+{ #category : #initialization }
+DRInterpreterToCompilerTest >> testGenerateSinglePrimitiveCompilationUnit [
+
+	| interpreterCompiler |
+
+	interpreterCompiler := DRInterpreterCompiler fromInterpreter: (self interpreterWith: #(#primitiveAdd)).
+	self assert: interpreterCompiler primitives notEmpty
```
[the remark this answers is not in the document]
hernanmd: These are DRPrimitiveObject. I will commit an better version.

Druid/CompiledMethod.extension.st
hernanmd: I don't know, I've got lost with "this", "them", "these".... Do you have an example?

Druid/Symbol.extension.st
hernanmd: To understand how it works for primitives, it is best to debug from #addPrimitivesTo: (this could be triggered sending #generateIR to an DRInterpreterCompiler) . #addPrimitivesTo: walks through the #interpreterPrimitiveTable, where each primitive is specified by a symbol, e.g. #primitiveRemLargeIntegers. A CompilationUnit instance then adds a new primitive (DRPrimitiveObject) identified with this Symbol (we store this in #sourceSelector:).  When we want to get the number of arguments of this primitive specified by the symbol we do:

```smalltalk
(self compilerCompiler numberOfArgumentsFor: aString)
```

(Ideally, the PrimitiveCompiler should have no problem resolving the number of arguments of a method, whether it is a name of a primitive or a CompiledMethod, for this there is the double dispatch).

We arrive at this method which returns the number of arguments of the primitive method:

```smalltalk
CompiledMethod>>numberOfArguments

	^ (self ast pragmaNamed: #numberOfArguments:) arguments first value
```

When the primitive does not have the numberOfArguments: pragma, the method that tries to get it (#pragmaNamed:) fails here:

```smalltalk
pragmaNamed: aSymbol
	^ self pragmaNamed: aSymbol ifAbsent: [ KeyNotFound signalFor: aSymbol  ]
```

So the issue is what we do with primitive without the pragma `<numberOfArguments: >`

Druid/OrderedCollection.extension.st
hernanmd: Messirve

Druid/Symbol.extension.st
hernanmd: For now I will implement:
```smalltalk
DRPrimitiveCompilerCompiler>>numberOfArgumentsFor: anObject
	"Answer a <Number> representing the number of arguments found in anObject (method or selector in the receiver's primitive's source)"

	^ [ anObject numberOfArgumentsIn: self ]
	on: KeyNotFound 
	do: [ : ex | ex return: 0 ]
```

Druid/DRPrimitiveCompilerCompilerCollector.class.st
hernanmd: Why couldn't be a subclass of DRPrimitiveCompilerCompiler if reuses behavior?

Druid/DRInterpreterBuilder.class.st
hernanmd: How would you group them?

Druid-Tests/DRPrimitiveCompilerCompilerCollectorTest.class.st
```diff
@@ -0,0 +1,121 @@
+Class {
+	#name : #DRPrimitiveCompilerCompilerCollectorTest,
```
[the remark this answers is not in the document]
hernanmd: The difference was a setUp method was missing and some copied methods weren't removed.
Fixed in the last commit

Druid/Symbol.extension.st
hernanmd: I will check Resolved and we could discuss later what we do with primitives without <numberOfArguments: >

PalumboN:
Can we merge this now? 
CI is failing on macOS (fail to compile Unicorn (?) but I cannot do much because I don't have powers in this repo 👎

Druid/CCodeGenerator.extension.st
hernanmd: Ok done

Druid/DRCompilationBuildUnit.class.st
hernanmd: Done

Druid/VMMaker.extension.st
hernanmd: I moved this PR to the pharo-vm

Druid/CCodeGenerator.extension.st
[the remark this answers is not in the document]
hernanmd: Moved to pharo-vm druid branch

guillep (dismissed):

guillep (approved):

PR #22, opened 2023-03-17
Fix druid compilation

PalumboN:
This PR fix mainly the `rawFloatBits` translation changed in https://github.com/pharo-project/pharo-vm/pull/548

It also  
- adds a script to create the Cogit class for Slang `DruidJIT` (?
- and remove the extension method `stopOnErrors` that now I don't know if this is ok or not 🤔


PalumboN:
Failing tests are because of ephemerons. I'm checking it.

Druid-Tests/DruidTestInterpreter.class.st
```diff
@@ -718,7 +718,7 @@ DruidTestInterpreter >> primitiveFloat64At [
 	addressToRead := self stackTop.
 	self
 		pop: 1
-		thenPush: (objectMemory float64At: addressToRead) rawFloatBits
```
guillep, line 721: Mea culpa, I introduced that in the druid experimental branch but never fixed it in slang.

I prefer that floats understand `rawFloatBits` instead of using  `rawFloatBitsOf:...`. What do you think?
Obtaining the bits of a float is something that is independent of the representation of objects, so ideally it should not be in the objectMemory

Druid-Tests/DruidTestInterpreter.class.st
```diff
@@ -734,58 +734,54 @@ DruidTestInterpreter >> primitiveFloat64AtPut [
...
 	<numberOfArguments: 1>
-	self
-		pop: 2
-		thenPush:
-		(self stackTop asFloat + (self stackValue: 1) asFloat) rawFloatBits
+	self pop: 2 thenPush: (objectMemory
+			 rawFloatBitsOf:
+			 self stackTop asFloat + (self stackValue: 1) asFloat
+			 in: nil)
 ]
```
guillep, line 746: You see, this reads much better (and the version on the right has that strange `nil` argument that seems unused)a...

guillep (approved):
I think it's ok to make it work, but I'd like to see back my `rawFloatBits` sent to a float instance :(

PalumboN:
@guillep Yes I moved because of the Slang, it was written in [the interpreter's definition using `self cCode:`](https://github.com/pharo-project/pharo-vm/pull/548/files#diff-7767e75b2e4228b2efd837d31440e1d3f4b10e494676b2c91c7c663dbc89d6dcL630) and it was the simplest way. 
But yes, I liked the other way, I'll open an issue.

PalumboN:
And I fixed the problem in the CI here: https://github.com/pharo-project/pharo-llvmDisassembler/pull/10

guillep:
> And I fixed the problem in the CI here: [pharo-project/pharo-llvmDisassembler#10](https://github.com/pharo-project/pharo-llvmDisassembler/pull/10)

Ok, so this PR will not work before the other is integrated. gotcha

Druid-Tests/DruidTestInterpreter.class.st
PalumboN: https://github.com/Alamvic/druid/issues/23

PalumboN:
Je ok, now it's failing because of a "bad fix" from a deprecation:

`Object >> asOrderedCollection` method is deprecated in Pharo 11 -> https://github.com/pharo-project/pharo/blob/Pharo11/src/Deprecated11/Object.extension.st#L4

But the rule for rewrite only works for objects that are not collections, but it is applied to parameters that can be collections (specifically one in PCG package). It changes:
```
anObject asOrderedCollection "When anObject is a collection, it returns a new one with the same elements"
-----
OrderedCollection wiith: anObject "When anObject is a collection, it returns a new collection with the collection as unique element"
```


@guillep  Is there a way to avoid deprecation rules in the CI?

guillep:
Ah I see it is in `PCGMessageNode>>#arguments:` ... @hernanmd maybe can help us here.
PCG is not being maintained. So either we maintain it, or we drop it...

PalumboN:
Yes, but I think anyway that the transformation has a bug, because the message is still valid for collections

guillep:
The problem is that the call site was probably polymorphic.

So there was a callsite that did `xxx asOrderedCollection` and `xxx` was not a collection. So the deprecation expanded it to: `OrderedCollection with: xxx`.

But then, later, `xxx` was an ordered collection (for example), and instead of the correct implementation  `asOrderedCollection ^self` we have `OrderedCollection with: aCollection`, giving us a collection of collections.

guillep (commented):
Wrap also the outer message node

PR #25, opened 2023-03-27
Type system for inline

PalumboN:
I extracted the logic of types inference for inlining in another object to play with POLYMORPHISM.

I needed this step to continue with my experiment building DROpal, one optimised bytecode compiler 🤓


PalumboN:
Yes! They are (?

PR #27, opened 2023-04-03
Task126 split2

hernanmd:
This PR includes changes to add an Interpreter Builder which can generate a narrowed version of an AbstractInterpreter (or StackInterpreter) with the same protocol as the Interpreter To Compiler for JIT compiler. The main difference is the Interpreter Builder does not compile any primitive, but generates methods to initialize its primitive table.

### DIFF
--- a/Druid/DRStackCompilationUnit.class.st
+++ b/Druid/DRStackCompilationUnit.class.st
@@ -1,0 +1,81 @@
+Class {
+	#name : #DRStackCompilationUnit,
+	#superclass : #DRPrimitiveCompilationUnit,
+	#instVars : [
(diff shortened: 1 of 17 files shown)


Druid/DRStackCompilationUnit.class.st
```diff
@@ -0,0 +1,81 @@
+Class {
+	#name : #DRStackCompilationUnit,
```
guillep, line 2: What is the objective of this class?

guillep (approved):
Let's wait for the CI to run!

guillep:
ci is green !

PR #33, opened 2023-04-06
Fix maycallback primitive selection

guillep:
Fix the selection of primitives so the ones as maycallback are listed but not generated.
Also, 
- Fixes in bytecode compilation (do not add extra return)
- some cleanups


Druid/DRInterpreterInstruction.class.st
```diff
@@ -24,6 +25,14 @@ DRInterpreterInstruction >> cfg: aDRPrimitiveControlFlowGraph [
...
 ]
 
+{ #category : #accessing }
+DRInterpreterInstruction >> genSelector [
+	" Answer a <Symbol> specifying the JITed selector of the interpreter's counterpart (interpreter) method"
+
+	^ genSelector
+		ifNil: [ genSelector := #gen_ , self sourceSelector capitalized ]
+]
+
```
PalumboN, line 35: I removed the variable. Duplicated code in subclass' override

PalumboN (approved):

PalumboN (approved):

guillep:
Gracias por le fix

PR #45, opened 2023-04-24
Suffix bytecode generators when required

guillep:
Expand bytecode entries if required.
Each entry should have a suffix.

Compile a method per suffixed variant


PalumboN (approved):
Cool!! Just a comment about the tests cases 👍

PR #46, opened 2023-04-25
Fix/bytecodeselection

guillep:
Depends on #45 

Make bytecode selection happen on the bytecode object, not on the compiled method.
This allows bytecode selection grab more information than just the source method's metadata.
For example, it has now where it is installed in the table, and it can have in the future information about its compilation


PalumboN:
There are some related tests broken. And there are changes already merged in `main`.

PR #50, opened 2023-04-25
Fix register allocation determinism

guillep:
Put some order in some iterations to disambiguate non determinism


PalumboN (approved):
Thank you 🙏

PR #54, opened 2023-05-24
Cogit: Object reference annotations 

PalumboN:
Related to https://gitlab.inria.fr/RMOD/alamvic/druid/-/issues/160

There are two things in this PR

### Annotate object references

Not  `#genConstant:R:` is used **always** that a constant is moved to a register. I don't know if it is the best, but that method has an if to know if it is an object reference or not, so no problems for now... I think.

And I needed to add a new copy to avoid the JITCompilerExpression transformation. It was the simplest way, controlling the SCCP for values (not instructions) is not easy (too many ifs).

### Loads in compile time

There is also something about moving LOADs from execution to compile time, something that we were talking about.

But now, not mixing the `methodObj` reference, is not used in any place. So let's see how it evolves.


Druid-Tests/DruidTestRTLCompiler.class.st
```diff
@@ -3260,24 +3249,21 @@ DruidTestRTLCompiler >> gen_pushLiteralConstantBytecode [
 	"AutoGenerated by Druid"
 
 	| currentBlock |
-	self MoveCq: ((methodObj + 8) + ((1 + LiteralStart) << 3)) R: TempReg.
-	self MoveM64: 0 r: TempReg R: TempReg.
-	self ssPushRegister: TempReg.
+	self genMoveConstant: methodObj R: TempReg.
+	self ssPushBase: TempReg offset: 1 + LiteralStart << 3 + 8.
 	^ 0
```
PalumboN, line 3254: Is it the expected behaviour?

guillep: Yes, this is many super very hyper recontra great

Druid-Tests/DRSCCPConstantFoldingTest.class.st
```diff
@@ -219,6 +219,41 @@ DRSCCPConstantFoldingTest >> testConstantFoldingFalseOrRegisterIsRegister [
...
+{ #category : #arithmetic }
+DRSCCPConstantFoldingTest >> testConstantFoldingLoadConstant [
+
+	| cfg |
+	cfg := self testConstantFoldingOf: DRLoad between: 8 asDRValue and: 9999 asDRValue.
+
+	self assert: cfg instructions allButLast last isCopy.
+	self
+		assert: cfg instructions allButLast last operand1 expression
+		equals: '(coInterpreter int64AtPointer:  9999)'
```
guillep, line 231: Nice! Now what would be important for a next iteration is to decide what is *constant at JIT compile time* and what it's not. Because if the value at address 9999 never changes it is constant and we can load it at compile time. Otherwise, we need to defer that to runtime.

Druid-Tests/DruidTestRTLCompiler.class.st
```diff
@@ -3260,24 +3249,21 @@ DruidTestRTLCompiler >> gen_pushLiteralConstantBytecode [
 	"AutoGenerated by Druid"
 
 	| currentBlock |
-	self MoveCq: ((methodObj + 8) + ((1 + LiteralStart) << 3)) R: TempReg.
-	self MoveM64: 0 r: TempReg R: TempReg.
-	self ssPushRegister: TempReg.
+	self genMoveConstant: methodObj R: TempReg.
+	self ssPushBase: TempReg offset: 1 + LiteralStart << 3 + 8.
```
guillep, line 3253: Lovely

Druid/DRBaseOffsetMemoryAddress.class.st
```diff
@@ -70,3 +70,9 @@ DRBaseOffsetMemoryAddress >> rtlArgumentExpressions [
 
 	^ { offsetNode . baseNode }
 ]
+
+{ #category : #SCCP }
+DRBaseOffsetMemoryAddress >> sccpLatticeValueFor: aDRSCCP [ 
+
+	^ aDRSCCP bottom "?"
```
guillep, line 77: I think maybe everything should be bottom by default, right? Instead of putting it in each valuable.

Now, I would leave here the bottom, maybe without the `"?"` comment. I don't see the value of constant propagating an address, because most of the time (always?) they are just operands of load and store instructions.

Druid/DRBytecodeIRGenerator.class.st
```diff
@@ -107,7 +107,7 @@ DRBytecodeIRGenerator >> interpretIframeMethodWith: aRBMessageNode [
 
 	^ self
 		  addInstructionFrom: aRBMessageNode
-		  instructionKind: DRCopy
+		  instructionKind: DRObjectReferenceCopy
```
guillep, line 110: 👍

guillep (approved):
Some comments along the way.

I'm going to integrate it, this is a lot of good work. Then I'm going to test it compiling the missing bytecodes we had :)

Druid/DRCogitStackToRegisterMappingGenerator.class.st
```diff
@@ -28,6 +28,17 @@ DRCogitStackToRegisterMappingGenerator >> returnNodeForCompiledMethod [
 	^ RBLiteralNode value: 0
 ]
 
+{ #category : #visiting }
+DRCogitStackToRegisterMappingGenerator >> twoOperandMoveCogitRTL: anInstruction operands: operands [
+
+	(self selectorCogitRTL: '' operands: operands instruction: anInstruction) = #Cq:R: ifTrue: [ 
+		"It's moving a constant to a register. This constant could be an oop, so we need to annotate it."
+		self flag: #TOTHINK. "Is this strategy ok?"
```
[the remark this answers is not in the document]
PalumboN: Yes, I know. 
Now that we have `DRObjectReferenceCopy` this logic should be attached to that. (It did not exist when I wrote this code).
But I was not 100% sure about it, so we can open an issue and see it in the next iterations.

Druid/DRBaseOffsetMemoryAddress.class.st
PalumboN: Yes, we can push up this. I prefer to not have foldings by default.

Druid-Tests/DRProductionBytecodeTest.class.st
```diff
@@ -219,10 +217,13 @@ DRProductionBytecodeTest >> testBytecodePushLiteralConstant [
...
 		withArguments: #(  ).
-		
+
 	self
 		assert: machineSimulator receiverRegisterValue
-		equals: (memory integerObjectOf: 42)
+		equals: (memory integerObjectOf: 42).
+
+	self assert:
+		(cogit hasAnnotatedAbstractInstructions: cogit getIsObjectReference)
```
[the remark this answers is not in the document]
PalumboN: Yes, I couldn't make that test (bah, it will take me more time and I wanted to know if I was by the right way).
That test is more integral, I think complementary to this one.

guillep:
Right on, PR is green!

Issue #55, opened 2023-06-01
Bytecode register allocation


guillep:
Register allocation in bytecodes

 - we need to allocate a minimum set of registers for each bytecode
 - one way to get the minimum number of registers needed by a method is to do an allocation with an infinite pool of registers working as a stack.
 - Once we know how many registers are needed, we need in each bytecode to allocate them before hand and check
  -  they do not overlap
  - if there is no available register, we exit compilation (see pseudocode below)

	live := 0.

	t0 := self allocateRegNotConflictingWith: live ifNone: [ ^ ABANDON COMPILATION ].
	live := live | t0.
	
	t1 := self allocateRegNotConflictingWith: live ifNone: [ ^ ABANDON COMPILATION ].
	live := live | t1.
	
	t2 := self allocateRegNotConflictingWith: live ifNone: [ ^ ABANDON COMPILATION ].
	live := live | t2.

guillep:
One thing to note is that the TempRegister is always available as scratch register.
This means we can use it freely, but we cannot push it to the stack

PR #57, opened 2023-06-01
Add register pool

guillep:
Introduce the concept of register pools.
A fixed register pool has a limited number of registers.
When no more registers are available, the register allocator will spill the register into a stack slot.

An infinite register pool always has new registers.
However, it will reuse them when possible (not conflicting live ranges).
This leads to an almost optimal (minimal?) usage of registers, depending on the register allocation algorithm of course.

First part of #55

### DIFF
--- a/Druid/DRInfiniteRegisterPool.class.st
+++ b/Druid/DRInfiniteRegisterPool.class.st
@@ -1,0 +1,11 @@
+Class {
+	#name : #DRInfiniteRegisterPool,
+	#superclass : #Object,
+	#instVars : [
+		'integerRegisters',
+		'floatingPointRegisters',
+		'spillRegisters',
+		'nextId'
+	],
+	#category : #'Druid-LinearScanRegisterAllocation'
+}
--- a/Druid/DRInfiniteRegisterPool.class.st
+++ b/Druid/DRInfiniteRegisterPool.class.st
@@ -4,8 +4,91 @@ Class {
 	#instVars : [
 		'integerRegisters',
 		'floatingPointRegisters',
-		'spillRegisters',
 		'nextId'
 	],
 	#category : #'Druid-LinearScanRegisterAllocation'
 }
+
+{ #category : #allocation }
+DRInfiniteRegisterPool >> createNewRegister [
+
+	| newRegister |
+	newRegister := 'T', nextId asString.
+	nextId := nextId + 1.
+	^ DRPhysicalGeneralPurposeRegister name: newRegister
+]
+
+{ #category : #initialization }
+DRInfiniteRegisterPool >> initialize [
+
+	super initialize.
+	nextId := 0.
+	
+	integerRegisters := Stack new.
+	floatingPointRegisters := Stack new.
+]
+
+{ #category : #allocation }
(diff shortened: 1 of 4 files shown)


Druid/DRCogitStackToRegisterMappingGenerator.class.st
```diff
@@ -68,6 +68,13 @@ DRCogitStackToRegisterMappingGenerator >> returnNodeForCompiledMethod [
 	^ RBLiteralNode value: 0
 ]
 
+{ #category : #visiting }
+DRCogitStackToRegisterMappingGenerator >> rtlExpressionForValue: aValue [
+
+	variables at: aValue result ifPresent: [ :var | ^ RBVariableNode named: var ].
+	^ super rtlExpressionForValue: aValue
+]
+
```
PalumboN, line 77: Why sometimes the order of the methods change? Could be a _non-deterministic situation_ in.... Fuel/Iceberg?

Druid/DRInfiniteRegisterPool.class.st
```diff
@@ -0,0 +1,94 @@
...
+	currentRegisters := aCollection copy.
+	inactiveOverlappingRegisters do: [ :e |
+		currentRegisters
+			remove: e
+			ifAbsent: [ "not care, we just want to ignore that register..." ] ].
+
+	"If we have registers, take one"
+	currentRegisters ifNotEmpty: [
+		presentBlock value: currentRegisters first ]
+]
```
PalumboN, line 94: This method is the same for Fixed and Infinite pools.
There are also parts of logic in common between the callers.

Maybe is worth pushing up to a superclass.... maybe not :)

Druid/DRLinearScanRegisterAllocator.class.st
```diff
@@ -4,47 +4,40 @@ Class {
...
+		ifTrue: [
+			registerPool
+				takeFloatingPointRegisterNotIn: inactiveOverlappingRegisters
+				ifPresent: [ :selectedRegister |
+				^ presentBlock value: selectedRegister ] ]
+		ifFalse: [
+			registerPool
+				takeIntegerRegisterNotIn: inactiveOverlappingRegisters
+				ifPresent: [ :selectedRegister |
+				^ presentBlock value: selectedRegister ] ]
```
PalumboN, line 40: Nice! I like more this version 👍

PalumboN (approved):
Piolaaaaaa 🚀 

I left one comment that you can take for the next iteration

Druid/DRCogitStackToRegisterMappingGenerator.class.st
guillep: tonel

Druid/DRInfiniteRegisterPool.class.st
guillep: yes, I prototyped it and I forgot

Druid/DRLinearScanRegisterAllocator.class.st
guillep: It can still be improved but yes

PR #59, opened 2023-06-02
Keep track of live registers during bytecode meta-compilation

guillep:
Keep track of live registers during dynamic allocation in bytecode compilation.

Requires changes in the latest VM to detect the failure to allocate registers.

Part of #55


guillep:
If you wait for me, next week I can add tests.

In any case, the CI will not pass without my changes in the VM :)

Druid/DRCogitStackToRegisterMappingGenerator.class.st
```diff
@@ -12,24 +12,44 @@ Class {
 DRCogitStackToRegisterMappingGenerator >> allocateVariable: aVariableName [
```
PalumboN, line 12: What is `aVariableName`? I think it is `aDRResult`. In that case, we could rename the variable... and the method.

PalumboN (dismissed):
Cooollll! 

Don't forget to make the PR with the changes in the VM.

And yes, tests 🚥

PalumboN (approved):

guillep:
Tests are green!

PR #61, opened 2023-06-05
2bytes bytecode compilation

PalumboN:
Compiling `extPushIntegerBytecode` :)


Druid-Tests/DRProductionBytecodeTest.class.st
```diff
@@ -87,6 +87,32 @@ DRProductionBytecodeTest >> testBytecodeDuplicateTop [
...
+
+	self
+		executeMethodWithReceiver: (memory integerObjectOf: 17)
+		withArguments: #(  ).
+
+	self
+		assert: machineSimulator receiverRegisterValue
+		equals: (memory integerObjectOf: 256 + 13).
+
+	self assert: cogit extB equals: 0
```
guillep, line 113: Should we also test the numExtB = 0 ?

guillep (approved):
Good!

Druid-Tests/DRProductionBytecodeTest.class.st
PalumboN: Could be... I don't know because is not used... Maybe we can clean it from the interpreter code.

PR #62, opened 2023-06-07
Cleanup/branches

guillep:
Collapse the implementation of DRBranchIfTrue into DRBranchIfCondition

Coauthored with @hernanmd and @Ducasse


Druid/DRBranchIfCondition.class.st
```diff
@@ -25,31 +45,127 @@ DRBranchIfCondition >> condition: aCondition [
...
 DRBranchIfCondition >> isBranchIfCondition [ 
 
 	^ true
 ]
 
+{ #category : #testing }
+DRBranchIfCondition >> isConditionalBranch [
+
+	^ true
+]
```
PalumboN, line 97: Now we can merge this, right?

PalumboN (dismissed):
Coollllll 🚀

Druid/DRBaseOffsetMemoryAddress.class.st
```diff
@@ -31,7 +31,7 @@ DRBaseOffsetMemoryAddress >> base: anObject [
 { #category : #printing }
 DRBaseOffsetMemoryAddress >> name [
 	
-	^ '[', base result name, ', offset=', offset name, ']'
+	^ '[', base result name, ', offset=', offset asDRValue name, ']'
```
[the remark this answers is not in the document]
guillep: I don't know, when inspecting the CFG we found a case where this broke in the roassal visualisation, and I patched it to avoid the tons of debuggers

Druid/DRBranchIfCondition.class.st
guillep: true

Druid/DRBranchCollapse.class.st
```diff
@@ -26,7 +26,9 @@ DRBranchCollapse >> applyTo: aDRControlFlowGraph [
 DRBranchCollapse >> canBeAppliedIn: aDRControlFlowGraph [
 
 	aDRControlFlowGraph blocksDo: [ :b | 
-		(b endInstruction isBranchIfTrue and: [ b endInstruction operand1 isSetConditionCode ]) ifTrue: [ 
+		(b endInstruction condition isEqualsThanComparison and: [ 
+			b endInstruction operand1 isSetConditionCode
+				and: [ b endInstruction operand2 = true asDRValue ] ]) ifTrue: [
```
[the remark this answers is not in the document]
guillep: I put the same condition in both

Druid/DRCPSEdge.class.st
```diff
@@ -94,7 +94,9 @@ DRCPSEdge >> integerOperation: anOperantion [
 
 { #category : #testing }
 DRCPSEdge >> isDead [
-	^ constraints anySatisfy: [ :c | c isEmpty ]
+
+	^ constraints notEmpty and: [
+		  constraints allSatisfy: [ :c | c isEmpty ] ]
```
guillep, line 99: @PalumboN please take a look at this line

PalumboN: Mmmm, I have doubts about it. If I remember well:

- One edge (the connection between 2 blocks) has the info about all possible values (or constraints) for each register at some point in the program (after the `source` block) inside a path.
- If one register has no possible values (is empty) then this edge is dead because exist one register without possible values for it.
- Could be other registers (for example constants copies) that will never be empty.

But, strangely, all tests are green... maybe some case for this (mixing dead and not dead) is missing...

Druid-Tests/DruidTestRTLCompiler.class.st
```diff
@@ -245,15 +245,26 @@ DruidTestRTLCompiler >> gen_primitiveAdd [
...
+	jumpNext := self Jump: 0.
+	jumpTrue jmpTarget: self Label.
+	self MoveCq: 1 R: TempReg.
+	jumpNext jmpTarget: self Label.
 	currentBlock := self Label.
-	jump1 jmpTarget: currentBlock.
+	jump2 jmpTarget: currentBlock.
+	self CmpCq: 1 R: TempReg.
+	jump2 := self JumpZero: 0.
 	self MoveCq: 10 R: ReceiverResultReg.
```
PalumboN, line 268: Now this code is longer, maybe is related to some `DeadBranchElimination` that we lost?

PalumboN (approved):
If you want we can merge it and open an issue for the broken Dead Path detection?

guillep:
👍

PR #69, opened 2023-06-26
refactoring-inline-optimization

tesonep:
- Extracting inline logic from the DRMessageSend to the DRInline optimization

- Setting the typeSystem in the irGenerator of the compilerCompiler

### DIFF
--- a/Druid/DRMessageSend.class.st
+++ b/Druid/DRMessageSend.class.st
@@ -3,8 +3,7 @@ Class {
 	#superclass : #DRInstruction,
 	#instVars : [
 		'methodNode',
-		'irGenerator',
-		'typeSystem'
+		'inlineGenerator'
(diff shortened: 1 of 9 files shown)


Druid/DRMessageSend.class.st
```diff
@@ -3,8 +3,7 @@ Class {
 	#superclass : #DRInstruction,
 	#instVars : [
 		'methodNode',
-		'irGenerator',
-		'typeSystem'
+		'inlineGenerator'
```
PalumboN, line 6: We need to think about how to move this outside the instruction.
I think that it is ok for now, we still can execute an inlining on one specific instruction (this is important for debugging)

PalumboN (approved):
Cooollll! Why you want to do this?

PR #71, opened 2023-06-30
Support conditional branch generation

guillep:
- marking deadCode after an unconditional branch
- unmarking it in case there could be a branch targetting the following code

- support for marking calls as mapped instructions to map at the return point


Druid-Tests/DruidTestRTLCompiler.class.st
```diff
@@ -3424,7 +3424,7 @@ DruidTestRTLCompiler >> gen_pushLiteralVariable16CasesBytecode [
 		      ifNone: [ ^ self unknownBytecode ].
 	live := live bitOr: (self registerMaskFor: t0).
 	self genMoveConstant: methodObj R: t0.
-	self MoveM64: 0 + LiteralStart << 3 + 8 r: t0 R: t0.
+	self MoveM64: 15 + LiteralStart << 3 + 8 r: t0 R: t0.
```
PalumboN, line 3427: This change is because all cases are tested using the same `gen` method, right? 🤔

Druid-Tests/DruidTestRTLCompiler.class.st
```diff
@@ -3628,20 +3628,22 @@ DruidTestRTLCompiler >> gen_shortConditionalJumpFalse [
...
 	live := live bitOr: (self registerMaskFor: t0).
-	(self ssValue: 0) moveToReg: t0.
+	(self ssValue: 0) copyToReg: t0.
 	self ssPop: 1.
 	self CmpCq: objectMemory falseObject R: t0.
 	jump1 := self JumpZero: 0.
 	self CmpCq: objectMemory trueObject R: t0.
 	jump2 := self JumpZero: 0.
 	self CallRT: ceSendMustBeBooleanTrampoline.
+	self annotateBytecode: self Label.
```
PalumboN, line 3638: What does it mean?

Druid/DRCogitStackToRegisterMappingGenerator.class.st
```diff
@@ -171,18 +184,28 @@ DRCogitStackToRegisterMappingGenerator >> visitBytecodeJump: aDRBytecodeJump [
 			 receiver: RBVariableNode selfNode
 			 selector: #genJumpTo:
 			 arguments:
-			 { (self jitCompileTimeExpression: aDRBytecodeJump operand1) })
+			 { (self jitCompileTimeExpression: aDRBytecodeJump operand1) }).
+	
+	"Everything that follows an unconditional jump is potentially dead code"
+	markDeadCode := true
```
PalumboN, line 190: Was this needed or it's an optimisation? And it is to support CFG with dead code or dead basic blocks? Because an optimised CFG should not have dead code :P

Druid/DRInterpreterToCompiler.class.st
```diff
@@ -83,6 +83,14 @@ DRInterpreterToCompiler class >> generateDruidJITModel [
...
 		"sendLiteralSelector2ArgsBytecode"
 			160 161 162 163 164 165 166 167 168
+		"shortUnconditionalJump"
+			176 177 178 179 180 181 182 183
+		"shortConditionalJumpTrue"
+			184 185 186 187 188 189 190 191
+		"shortConditionalJumpFalse"
+			192 193 194 195 196 197 198 199
+		"popStackBytecode"
+			216
```
PalumboN, line 93: 👏

PalumboN (dismissed):
Niceee!!! 👏 

I left some questions. 

Tests are failing a lot, I think we need to merge the VM before: https://github.com/pharo-project/pharo-vm/pull/635

Druid-Tests/DruidTestRTLCompiler.class.st
guillep: yup

Druid-Tests/DruidTestRTLCompiler.class.st
guillep: This means that the mapping between machine code => bytecode

Druid/DRCogitStackToRegisterMappingGenerator.class.st
guillep: This is needed for cogit (It's in DRCogitStackToRegisterMappingGenerator). Cogit has a internal flag saying that the subsequent code is dead. And that flag is set to true when compiling an unconditional jump.
However, if the subsequent instructions are not dead, we need to mark them back as not dead code.

PalumboN:
After merging the PR at the VM the CI is still failing. 
```
Instance of CogSimStackEntry did not understand #popToReg:
```

PalumboN:
VAMOOOOOO 🚀

PR #82, opened 2023-11-14
Store and pop receiver variable bytecode

PalumboN:
Adding supports for bytecodes `200 201 202 203 204 205 206 207`

## Merge PR in the VM before

https://github.com/pharo-project/pharo-vm/pull/711

### DIFF
--- a/Druid-Tests/DRProductionBytecodeTest.class.st
+++ b/Druid-Tests/DRProductionBytecodeTest.class.st
@@ -1477,6 +1477,42 @@ DRProductionBytecodeTest >> testStoreAndPopReceiverVariableBytecode: index [
 ]
 
 { #category : #tests }
+DRProductionBytecodeTest >> testStoreAndPopReceiverVariableBytecodeCallRememberTrampoline [
+
+	| object value storeCheckTrampoline |
+
+	storeCheckTrampoline := self compile: [ cogit RetN: 0 ].
+	cogit ceStoreCheckTrampoline: storeCheckTrampoline.
+
+	self
+		compileBytecode: 200
+		selector: #storeAndPopReceiverVariableBytecode
+		thenDo: [ :generator |
+			cogit ssPushRegister: TempReg.
+
+			"Execute the druid's compiled code"
+			generator value.
+
+			"Then return without druid's compiled code"
+			cogit genUpArrowReturn ].
+
+	object := self newOldSpaceObjectWithSlots: 1.
+	value := self newObjectWithSlots: 0.
+
+	machineSimulator temporaryRegisterValue: value.
+
+	self
+		prepareStackForPrimitiveReceiver: object
+		arguments: #(  )
+		method: 0.
+
+	"Should arrive to trampoline to put the object in the remembered set"
+	self runFrom: cogInitialAddress until: storeCheckTrampoline.
+	
+	self assert: machineSimulator receiverRegisterValue equals: object
+]
+
+{ #category : #tests }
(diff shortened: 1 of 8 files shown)


Druid-Tests/DRProductionBytecodeTest.class.st
```diff
@@ -1436,3 +1435,141 @@ DRProductionBytecodeTest >> testReturnTopFromMethod [
...
+
+			"Execute the druid's compiled code"
+			generator value.
+
+			"Then return without druid's compiled code"
+			cogit genUpArrowReturn ].
+
+	object := self newObjectWithSlots: index + 1.
+
+	machineSimulator temporaryRegisterValue: memory trueObject.
```
guillep, line 1472: This goes to the remembered set in our case right?

Druid/DRBytecodeIRGenerator.class.st
```diff
@@ -303,6 +305,27 @@ DRBytecodeIRGenerator >> interpretPushWith: aRBMessageNode [
...
+		   addInstructionWithNoResultFrom: aRBMessageNode
+		   instructionKind: DRCall
+		   operands: {
+				   'ceStoreCheckTrampoline'.
+				   obj }) beMapped saveLinkReg
+
+	"	
+	cogit 
+		evaluateTrampolineCallBlock: [cogit CallRT: ceStoreCheckTrampoline]
+		protectLinkRegIfNot: inFrame.
```
guillep, line 325: Ok, we should review this later, I wonder why we need a special trampoline call.

Druid/DRIRGenerator.class.st
```diff
@@ -311,6 +311,8 @@ DRIRGenerator >> initializeSpecialCases [
 	
 	
 	specialCases at: #longAt:put: put: #interpretMachineWordAtPutWith:.
+	specialCases at: #unsignedLong64At:put: put: #interpretMachineWordAtPutWith:.
+	specialCases at: #unsignedLongAt:put: put: #interpretMachineWordAtPutWith:.
```
guillep, line 315: Ah we should do a pass on all the loads and stores after all the branches are integrated :)

Druid/DRIRGenerator.class.st
```diff
@@ -1806,6 +1808,8 @@ DRIRGenerator >> visitMessageNode: aRBMessageNode receiver: receiver [
 { #category : #visiting }
 DRIRGenerator >> visitMethodNode: aRBMethodNode [
 
+	(aRBMethodNode hasPragmaNamed: #doNotGenerate) ifTrue: [ 1 halt ].
```
guillep, line 1811: Interesting, have a case?
Why not the same pragma that in the other PR for the halt?

guillep (approved):
Good!

Druid-Tests/DRProductionBytecodeTest.class.st
PalumboN: Nop, the object is in the new space, nothing to remember.

Druid/DRBytecodeIRGenerator.class.st
PalumboN: Now we are generating something like that, I forget the comment.

Druid/DRIRGenerator.class.st
PalumboN: Hahahah I used it to know quickly where the meta-compilation scapes from the "real" interpreter code.

Good to find missing special cases. We can remove it (or not... should never happen I think 🤔 )

PR #84, opened 2023-11-16
Staged JIT expressions

PalumboN:
Fix on staged expression to compile `extB` bytecode :)

There are also:

- Some fixes for Constraints inference (with staged values)
- Avoid unnecessary Moves (due to staged expressions)

## TODO
- [x] Keep access to VMMemoryMap as JIT expressions
- [x] Compile a DruidVM that does not fail

## For next iteration
- [x] Fix `gen_ExtUnconditionalJump` -> `annotateBytecode:` https://github.com/Alamvic/druid/pull/86
- [ ] Fix write barrier in `storeAndPopReceiverVariableBytecode` https://github.com/Alamvic/druid/pull/87
- [x] Improve `generatePath` for JIT expressions (now are ignored, but registers can have constraints)
- [x] Fix staged loads
- [ ] Check why is it using more registers for some bytecodes (?

PR #86, opened 2023-11-28
Fix ext unconditional jump

PalumboN:
#### Missing work from https://github.com/Alamvic/druid/pull/84 

The problem was the missing `annotateBytecode` in some paths of the JIT bytecode definition. 
Activating the _staged compilation_ causes that sometimes the bytecode was not annotated, but it is marked as `isMapped` so the VM is crashing.

- Added an instruction for `annotateBytecode` at the beginning of a method.
- Avoid that for message sends (they use another annotation). New pragma was added to them. https://github.com/pharo-project/pharo-vm/pull/719
(That should be merged for the CI here)

### TODO
- [x] There is still one test skipped (extended backjump)


guillep (approved):
go!

PR #89, opened 2023-12-06
Linearize simulated stack

PalumboN:
Changes in the `reversePostOrderBlocks` algorithm to decide the order of conditional branches.

It is based on a new `stackDependency` for instructions, used to maintain the same order as the original code.

### DIFF
--- a/Druid/DRBytecodeIRGenerator.class.st
+++ b/Druid/DRBytecodeIRGenerator.class.st
@@ -523,6 +523,13 @@ DRBytecodeIRGenerator >> receiver: aDRValue [
 ]
 
 { #category : #visiting }
+DRBytecodeIRGenerator >> trackStackInstruction: aDRStrackInstruction [ 
+
+	aDRStrackInstruction stackDependency: executionState lastStackInstruction.
+	executionState lastStackInstruction: aDRStrackInstruction.
+]
+
+{ #category : #visiting }
 DRBytecodeIRGenerator >> visitClassVariableNode: aRBVariableNode [
 
 	| value mappings |
--- a/Druid/DRIRGenerator.class.st
+++ b/Druid/DRIRGenerator.class.st
@@ -42,11 +42,13 @@ DRIRGenerator >> addFrameReturn: aDRStackFrame [
 
 { #category : #visiting }
 DRIRGenerator >> addInstruction: instruction from: aNode [
-	 
+
 	instruction origin: thisContext sender stack.
 	self currentBasicBlock addInstruction: instruction.
 	instruction originAST: aNode.
 	self pushOperand: instruction.
+	instruction isStackInstruction ifTrue: [
+		self trackStackInstruction: instruction ].
 	^ instruction
 ]
 
@@ -1115,20 +1117,20 @@ DRIRGenerator >> interpretPopThenPushWith: aRBMessageNode [
 	Pop: keep track of the operand stack and remove one element
 	Push: move the value to the physical register assigned to the current stack position"
 	
-	| elementsToPop elementsToPopNode toPush |
+	| nElementsToPop elementToPush |
 	aRBMessageNode arguments first acceptVisitor: self.
-	elementsToPopNode := self popOperand.
-	elementsToPop := elementsToPopNode simpleConstantFold.
-	elementsToPop isInteger
+	nElementsToPop := self popOperand simpleConstantFold.
+	nElementsToPop isInteger
 		ifFalse: [ self error: 'expected an integer value' ].
 
 	"If it's a pop with push we should evaluate the pushed argument before popping in case it uses the stack elements"
 	aRBMessageNode arguments size > 1 ifTrue: [
 		aRBMessageNode arguments second acceptVisitor: self.
-		toPush := self popOperand ].
-	self pop: elementsToPop.
+		elementToPush := self popOperand ].
+	
+	self pop: nElementsToPop.
 	aRBMessageNode arguments size > 1 ifTrue: [
-		self push: toPush ].
+		self push: elementToPush ].
 
 	"pop:[thenPush:] does not return an interesting value"
 	self pushOperand: DRNullValue new.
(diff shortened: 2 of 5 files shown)


Druid/DRBasicBlock.class.st
```diff
@@ -197,6 +197,24 @@ DRBasicBlock >> breakBy: instruction [
...
+{ #category : #asserting }
+DRBasicBlock >> canBeLinearizedBefore: aDRBasicBlock [
+
+	self = aDRBasicBlock ifTrue: [ ^ true ].
+
+	"Domination constraints"
+	(self isDominatedBy: aDRBasicBlock) ifTrue: [ ^ false ].
+
+	"Pops and Pushes are constraints for linearization"
+	controlFlowGraph stackInstructions do: [ :stackInstruction | "There is a stack instruction in my path ..."
```
guillep, line 209: I wonder if we don't need to put this second part of the sorting in a separate (sub)class. This is only required to stage push/pop instructions of the abstract interpreter, which only happens in one of our backends.

Also, I wonder if this cannot just be rewritten a bit simpler as a topological sort with extra constraints.

Druid/DRBytecodeIRGenerator.class.st
```diff
@@ -522,6 +522,13 @@ DRBytecodeIRGenerator >> pushFrameForCode: aDRCode receiver: aReceiver [
 DRBytecodeIRGenerator >> receiver: aDRValue [
 ]
 
+{ #category : #visiting }
+DRBytecodeIRGenerator >> trackStackInstruction: aDRStrackInstruction [ 
+
+	aDRStrackInstruction stackDependency: executionState lastStackInstruction.
+	executionState lastStackInstruction: aDRStrackInstruction.
```
guillep, line 529: :D

Druid/DRCogitCodeGenerator.class.st
```diff
@@ -905,13 +905,6 @@ DRCogitCodeGenerator >> visitMultiplyOverflow: aDRMultiplyOverflow inBranch: aDR
 	^ aDRBranchIfCondition
 ]
 
-{ #category : #visiting }
-DRCogitCodeGenerator >> visitMultiplyOverflowJump: aDROverflowJump [ 
-	
-	self generateJump: #JumpMultiplyOverflow: target: aDROverflowJump trueBranch.
-	self generateJump: #JumpMultiplyNoOverflow: target: aDROverflowJump falseBranch.
-]
-
```
guillep, line 914: Why was this removed? STRANGEEE

Druid/DRIRGenerator.class.st
```diff
@@ -42,11 +42,13 @@ DRIRGenerator >> addFrameReturn: aDRStackFrame [
...
 { #category : #visiting }
 DRIRGenerator >> addInstruction: instruction from: aNode [
-	 
+
 	instruction origin: thisContext sender stack.
 	self currentBasicBlock addInstruction: instruction.
 	instruction originAST: aNode.
 	self pushOperand: instruction.
+	instruction isStackInstruction ifTrue: [
+		self trackStackInstruction: instruction ].
```
guillep, line 51: You're lazy xD You could have put this in the push and pop only, no?

Druid/DRMultiplyOverflow.class.st
```diff
@@ -16,16 +16,14 @@ DRMultiplyOverflow >> isMultiplyOverflow [
...
 	^ 'Multiply Overflow'
 ]
-
-{ #category : #printing }
-DRMultiplyOverflow >> reversed [
-
-	"Returns the same condition but with reversed operands: x = y --> y = x"
-1halt.
-	^ DREqualsThanComparison new
-]
```
guillep, line 31: dead codee :)

Druid/DROverflow.class.st
```diff
@@ -47,14 +47,6 @@ DROverflow >> prettyPrintWithBranch: aDRBranchIfCondition [
...
 ]
 
-{ #category : #printing }
-DROverflow >> reversed [
-
-	"Returns the same condition but with reversed operands: x = y --> y = x"
-1halt.
-	^ DREqualsThanComparison new
-]
-
```
guillep, line 57: 💀

guillep (approved):
Thanks!

Druid/DRBasicBlock.class.st
```diff
@@ -197,6 +197,24 @@ DRBasicBlock >> breakBy: instruction [
 		  self }
 ]
 
+{ #category : #asserting }
+DRBasicBlock >> canBeLinearizedBefore: aDRBasicBlock [
+
+	self = aDRBasicBlock ifTrue: [ ^ true ].
+
+	"Domination constraints"
+	(self isDominatedBy: aDRBasicBlock) ifTrue: [ ^ false ].
```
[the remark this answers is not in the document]
PalumboN: Yes, a conditional jump with a backjump in one branch and the loop exit in the other. 
Maybe we can avoid that construction, but I remember other problems in that scenario... we can check.

Also, now the dominator tree is built in many places, should be nice to have it cached.

Druid/DRCogitCodeGenerator.class.st
PalumboN: Because now all the logic is in [this awful method](https://github.com/Alamvic/druid/blob/9a5e22191191cd8fef108d2b18f6976576f32d96/Druid/DRCogitCodeGenerator.class.st#L44)

Druid/DRIRGenerator.class.st
PalumboN: I did it first, but here is better, because some Generators on interpreting pop and push don't create the instructions (just affect the VMState or something like that).

.... and the code was uglier 😨

Druid/DRCogitCodeGenerator.class.st
guillep: you are weak

Druid/DRBasicBlock.class.st
PalumboN: https://github.com/Alamvic/druid/issues/91

Druid/DRBasicBlock.class.st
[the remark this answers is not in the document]
PalumboN: https://github.com/Alamvic/druid/issues/91

PR #90, opened 2023-12-08
Stage loop

PalumboN:
Support for the generation of staged loop - i.e. generate `while` in cogit code.

- Changes in `generateCompileTimeIfCondition:` to select if compile a `while` or an `if`
- New "optimisation" `DRStoreTemporaryStagedLoop` (this is needed to generate the cogit code), to allocate the loop induction variables in temporaries.
- Some tests, not too much

### TODO

- [x] Test all conditional staged generation (if and while).
-


Druid/DRCogitCodeGenerator.class.st
```diff
@@ -311,6 +320,31 @@ DRCogitCodeGenerator >> generateJump: jumpKind target: aTarget [
...
+{ #category : #'code-generation' }
+DRCogitCodeGenerator >> generateLoop: aDRLoop [
+
+	| innerGenerator jitCondition loopPath isTrue |
+	loopPath := aDRLoop bodyBlocks.
+
+	self assert:
+		(loopPath allSatisfy: [ :b | b isDominatedBy: aDRLoop headerBlock ]).
+	blocks removeAll: loopPath.
+	innerGenerator := self innerCompileBlocks: loopPath.
```
guillep, line 332: Ok, here the variable name confused me at first.
This is not a loop path! This is all the blocks dominated by the loop header!

We could use the same for the ifs without having traces :')

Druid/DRCogitCodeGenerator.class.st
```diff
@@ -266,15 +266,24 @@ DRCogitCodeGenerator >> generateCompare: aComparison operands: operands [
...
-	sequenceNode := self innerCompilePathFrom: aDRBranchIfCondition trueBranch.
+
+]
+
+{ #category : #'code-generation' }
+DRCogitCodeGenerator >> generateConditional: aDRBranchIfCondition [
+
+	| jitCondition innerGenerator |
+	jitCondition := aDRBranchIfCondition conditionAsJitCompileTimeExpression.
+	innerGenerator := self innerCompilePathFrom: aDRBranchIfCondition trueBranch.
```
guillep, line 281: Crazy idea: what if we do:

```smalltalk
stagedTrueBranch := self innerCompileBlocks: aDRBranchIfCondition trueBranch.
stagedFalseBranch := self innerCompileBlocks: aDRBranchIfCondition falseBranch.
```

And then we generate an ifTrue:ifFalse: ?
Worst case, one of the branches will be an empty block, right?

Druid/DRCogitCodeGenerator.class.st
```diff
@@ -405,6 +439,45 @@ DRCogitCodeGenerator >> initialize [
...
+	dominatorTree := aDRBasicBlock controlFlowGraph buildDominatorTree.
+	dominatedBlocks := aDRBasicBlock controlFlowGraph reversePostOrderBlocks select: [ :b |
+		                   b isDominatedBy: aDRBasicBlock ].
+
+	"If the first block is dominated by others don't remove from the list, we still need it"
+	self assert: dominatedBlocks first = aDRBasicBlock.
+	(blocks includesAny: aDRBasicBlock predecessors) ifFalse: [ blocks remove: aDRBasicBlock ].
+
+	"Remove blocks only visited during inner compilation from outer"
+	dominatedBlocks allButFirstDo: [ :b | blocks remove: b ].
```
guillep, line 467: I don't see when this condition may happens

Druid/DRCogitCodeGenerator.class.st
```diff
@@ -831,11 +904,13 @@ DRCogitCodeGenerator >> visitIntToFloatConversion: aDRIntToFloat [
...
 { #category : #visiting }
-DRCogitCodeGenerator >> visitJump: aDRJump [ 
+DRCogitCodeGenerator >> visitJump: aDRJump [
 
-	self
-		generateJump: #Jump:
-		target: aDRJump target
+	aDRJump asBackJumpLoop ifNotNil: [ :loop |
+		loop loopConditional isJITCompileTimeExpression ifTrue: [
+			^ self "JIT compile loop" ] ].
```
guillep, line 911: Strangee

Druid/DRJump.class.st
```diff
@@ -21,6 +21,14 @@ DRJump >> acceptVisitor: aVisitor [
 	^ aVisitor visitJump: self
 ]
 
+{ #category : #converting }
+DRJump >> asBackJumpLoop [
+
+	^ self controlFlowGraph allLoops
```
guillep, line 27: strange too

Druid/DRLoop.class.st
```diff
@@ -171,6 +171,13 @@ DRLoop >> headerBlock [
 	^ self latchBlock successor.
 ]
 
+{ #category : #accessing }
+DRLoop >> inductionInitialValues [
+
+	^ self loopPhiFunctions flatCollect: [ :phi |
+		  phi valuesForPredecessor: self preheaderBlock ]
+]
```
guillep, line 179: like it :)

Druid/DRStoreTemporaryStagedLoop.class.st
```diff
@@ -0,0 +1,41 @@
...
+{ #category : #accessing }
+DRStoreTemporaryStagedLoop >> applyTo: cfg [
+
+	cfg phiFunctions
+		select: [ :phi | phi hasRecursiveUse and: [ phi isJITCompileTimeExpression ] ]
+		thenDo: [ :phi | self storeOnNewTemporaryVariables: phi ]
+]
+
+{ #category : #store }
+DRStoreTemporaryStagedLoop >> storeOnNewTemporaryVariables: phiFunction [
```
guillep, line 16: I don't understand what this does :)

guillep (approved):
Some comments on the way!

Druid/DRCogitCodeGenerator.class.st
PalumboN: It is not **all** the blocks dominated by the header, it is the _loopBody_ :P (or `loopBodyBlocks` to say that they could be many).
The assertion is to be safe in case of jumps to the middle of the body (I'm not sure what happens then).

The if compilation also does this to know if blocks are shared between the staged and not staged paths.

The traces are there to avoid some phi functions. I don't remember the problem but we can check that easily.

Druid/DRCogitCodeGenerator.class.st
PalumboN: The problem is not the empty block but the phis after the if.

Here you also need to calculate the path from each branch until the joint point.

I think that the problem includes the result of the phi as a jit expression. In that case, we need to store variable(s) similar to the inductive variables for the loop in this PR.

I'm not sure what is better...

Druid/DRCogitCodeGenerator.class.st
guillep: yes, it's not the ones dominated by the header. It's the ones dominated by the loop body block.
What we need is all the blocks up to the domination frontier.

Yes let's check :)

Druid/DRCogitCodeGenerator.class.st
PalumboN: Hahaha you don't want to generate the backjump, it is implicit in the `while`.

But maybe we should move all this JITCompile stuff to subclasses...

guillep: I think what I meant was that `asBackJump` returning nil is strange :)

Druid/DRJump.class.st
PalumboN: Hahah what? The name or the logic? I listen for suggestions :P

Druid/DRStoreTemporaryStagedLoop.class.st
PalumboN: image

Hahaha I will add a comment, but I want to improve this in the register allocator. 

Now it is "creating" the variables that iterate the loop, which are variables in the cogit method.
I tried to do it directly in the code generation but doing this is too much simple.

PR #93, opened 2023-12-14
Simple druid

guillep:
Add a new druid backend that outputs code for the simple stack cogit.
This leaves the same set of bytecodes and primitives working for the two cogit backends.
That is, there are still bugs with
- the staging of `pushLiteralVariable16CasesBytecode`
- the `extPushFullClosureBytecode` 

Depends on https://github.com/pharo-project/pharo-vm/pull/724

In detail ths PR does:
- add a simple cogit backend for both primitive and bytecode generaiton
- refactor tests to make them parametrisable with the new backends
- allow configuring the set of optimisations depending on the backend
- decouple a bit more the IR generator from the backend using the LoadReceiver and LoadArgument instructions
- add tests about signed negative comparisons
- add canonicaliser rules for push and copy instructions
- change the type of `DRLoadArgument`: ⚠️ this could break the bytecode generation backend
- add `DRReceiverConstraint` which adds constraints on all `DRLoadReceiver` instructions. Q for @PalumboN : is this required? Or knowing that we have an instruction with particular semantics (loads the receiver) how could we add the constraint to that?


Druid/DRSmallIntegerPrimitiveCustomisation.class.st
```diff
@@ -19,9 +19,7 @@ DRSmallIntegerPrimitiveCustomisation >> generateCustomisedPreamble: aGeneratorMe
 { #category : #'generation-IR' }
 DRSmallIntegerPrimitiveCustomisation >> initialConstraint [
 
-	^ DRRegisterDomain
-		  equals: (DRPhysicalGeneralPurposeRegister name: 'ReceiverResultReg')
-		  than: 1
+	^ DRReceiverConstraint with: (DREqualsConstraint withValue: 1)
```
PalumboN, line 22: Mmmm here you are changing a Domain by a Constraint, that is not good. A domain is a tuple of Register -> Constraint.

Druid/DRReceiverConstraint.class.st
```diff
@@ -0,0 +1,32 @@
+Class {
+	#name : #DRReceiverConstraint,
+	#superclass : #Object,
+	#instVars : [
+		'constraint'
+	],
```
PalumboN, line 6: Mmm a Constraint with a constraint inside? 🤔

Druid/DRReceiverConstraint.class.st
```diff
@@ -0,0 +1,32 @@
...
+{ #category : #adding }
+DRReceiverConstraint >> addToEdge: aDRCPSEdge [
+
+	| loadReceiverInstructions |
+	loadReceiverInstructions := aDRCPSEdge source controlFlowGraph
+		                            instructions select: [ :e |
+		                            e isLoadReceiver ].
+	loadReceiverInstructions do: [ :instruction |
+		aDRCPSEdge addConstraint: constraint to: instruction result name ]
+]
```
PalumboN, line 27: What do you want to do here? This doesn't seem good because the edge has info about the paths before, and here you are taking the info from all the CFG.

PalumboN: Ok, what we cannot say now is the register where is the receiver, right?

In that case, maybe we need to check what the visitor is doing with the DRLoadReceiver instruction (maybe now is doing nothing).
The problem is how to define the specialization in that case, we need to think about that.

PalumboN (dismissed):

Druid/DRReceiverConstraint.class.st
guillep: Anyways this does not seem to have bad side effects for now, as the LoadReceiver instruction is always at the top of the cfg

PR #94, opened 2023-12-15
Fix closure creation

PalumboN:
Fix the compilation of `gen_ExtPushFullClosureBytecode` 🥳 

- Saving fixed register used for the closure creation intrinsic
- Avoiding `= true` in conditions
- Allocate dynamic registers outside the loops and staged paths

DruidJIT -> https://github.com/pharo-project/pharo-vm/pull/725/


guillep (approved):
GOGOGO

PR #101, opened 2024-02-07
Compile primitive at

PalumboN:
`primitiveAt` compilation:

- Avoid some cases where both Druid compilation and/or Cogit compilation were failing.
- Change the return compilation code based on meta-interpretation (if there are ignored cases then the primitive is _incomplete_, aka returns 0).

### VM changes https://github.com/pharo-project/pharo-vm/pull/736


guillep (approved):
Looks good!

PR #113, opened 2024-02-22
Fix float support

guillep:
- Translations to get the internal representation of floats (and  viceversa)
- Make a real distinction between arithmetic and logical shift right
- Fix type of bitAnd: (should be unsigned)

+ tests and some cleanups on the way


Druid-Tests/SimpleDruidTestRTLCompiler.class.st
```diff
@@ -3785,19 +3785,13 @@ SimpleDruidTestRTLCompiler >> gen_primitiveLessThanValue [
...
-	self CmpCq: 15 R: SendNumArgsReg.
-	jump1 := self JumpGreaterOrEqual: 0.
-	self MoveR: SendNumArgsReg R: ClassReg.
-	jump2 := self Jump: b16.
-	currentBlock := self Label.
-	jump1 jmpTarget: currentBlock.
-	self MoveR: SendNumArgsReg R: ReceiverResultReg.
+	s6 := s3 + 1.
+	[ ((s6<15)) ] whileTrue: [ s3 := s6 ].
+	self MoveCq: s6 R: ReceiverResultReg.
```
guillep, line 3794: I don't think that this is the fault of this PR.
I ran the tests locally so this code was regenerated recently...

Druid-Tests/SimpleDruidTestRTLCompiler.class.st
```diff
@@ -3806,19 +3800,13 @@ SimpleDruidTestRTLCompiler >> gen_primitiveLoopIncrementing [
...
```
guillep, line 3809: Same here

guillep (commented):
Some investigations are needed with the loop staging

PR #115, opened 2024-03-04
Rotate instruction

doste:
Fix https://github.com/Alamvic/druid/issues/112

- Added a rotate instruction to druid's IR
- The Cogit code generator supports said instruction, it'ss mapped to the Cogit's RotateLeftCq:R: method
- Added the special case in DRIRGenerator and all the visit methods needed for the instruction to work

PR #116, opened 2024-03-06
Better canonicalisation

doste:


Druid/DRCogitCanonicaliser.class.st
```diff
@@ -10,6 +10,41 @@ DRCogitCanonicaliser >> applyTo: cfg [
 	cfg instructions copy do: [ :e | e acceptVisitor: self ]
 ]
 
+{ #category : #visiting }
+DRCogitCanonicaliser >> putConstantsTogetherInDoubleInstructionIfPossible: aDRBinaryInstruction [
+	"If it is a pattern such as res := ((var op const1) op const2)"
```
PalumboN, line 15: Now this method do more stuff.

I propose this comment:

```suggestion
DRCogitCanonicaliser >> putConstantsTogetherInDoubleInstructionIfPossible: aDRBinaryInstruction [
	"If it is a pattern such as ((var op const) op X)
	It is replaced by:
		((X op const) op var) -> if X is constant
		((var op X) op const) -> otherwise
	"
```

And maybe change the name to `sortOperandsWithContant...` or something like that.

Druid/DRCogitCanonicaliser.class.st
```diff
@@ -292,9 +276,23 @@ DRCogitCanonicaliser >> visitMod: aDRMod [
...
+			exponent := (aDRMultiply operand2 value log: 2) asInteger.
+			shiftInstr := DRLogicalLeftShift
+				              operands: {
+						              aDRMultiply operand1.
+						              exponent asDRValue }
+				              result: aDRMultiply result.
+
+			aDRMultiply replaceBy: shiftInstr ]
+		ifFalse: [ "Else, try to put the constants (or the variables) together"
+			self putConstantsTogetherInDoubleInstructionIfPossible: aDRMultiply ]
```
PalumboN, line 295: These cannot be both applied? 🤔 

Maybe:
```suggestion
	| exponent shiftInstr |
	self putConstantsTogetherInDoubleInstructionIfPossible: aDRMultiply
	self transformMultiplyByShiftIfPossible: aDRMultiply
```

PalumboN (commented):
Cool!! Good improvements! 🎈 

We should regenerate the DruidJIT and send a PR to the VM to use this on the benchmarks.

I left some comments to check.

### We should merge https://github.com/Alamvic/druid/pull/115 before

Druid-Tests/DruidTestRTLCompiler.class.st
```diff
@@ -2955,6 +2955,54 @@ DruidTestRTLCompiler >> gen_primitiveDivideByConstant [
...
+DruidTestRTLCompiler >> gen_primitiveDoubleBitAndSubtreeRightVariableRight [
+	"AutoGenerated by Druid"
+
+	| currentBlock |
+	self MoveR: ReceiverResultReg R: ClassReg.
+	self AndCq: 16r3FFFFF R: ClassReg.
+	self MoveR: ClassReg R: ReceiverResultReg.
+	self genPrimReturn.
+	^ CompletePrimitive
+]
```
PalumboN, line 3004: Is this right? (all the methods are equal). These are the generated, maybe it was missing to run the tests at the end?

PalumboN (approved):
Cool! I think that we should merge `main` here, now that we merged the other PR

PalumboN (approved):

PR #121, opened 2024-03-13
Fix full value

PalumboN:
Many fixes:

- Fix the primitive value for Intel architectures (TOCHECK)
- Fix the canonicalizer on sorting double operations
  - `DRInstruction >> isDominatedBy:` now also checks the position of the instruction inside the block
- Fix the rotates (type was missing and it was inferring signed ints)


doste (approved):
Good!

PR #124, opened 2024-03-14
Specialize primitiveAt based on format

PalumboN:
Here is the code used for the experiment of compiling specialized versions of the `primitiveAt`.

- I created a subclass `DRSpecialPrimitiveCompilerCompiler` with the magic
- It looks for the format number to specialize
https://github.com/PalumboN/opensmalltalk-vm/blob/503cb52076a8bda5ef88a8eb8862004accd80135/smalltalksrc/VMMaker/InterpreterPrimitives.class.st#L978-L987
- And adds a special case to interpret the format as a constant directly.


PalumboN:
We should move this experiment to evref org: https://github.com/evref-inria

PalumboN:
Moved to https://github.com/evref-inria/druid/pull/1

PR #127, opened 2024-03-15
Fix a few optimizations

doste:
Made a few fixes in the optimizations methods, mainly:

* isDominatedBy:
 This method is called when doing a Loop Invariant Code Motion pass and the way it was written before, it made a few tests fail because when doing this optimization a BasicBlock may have been duplicated and the dominance of the blocks was not set accordingly.

* sortOperandsInDoubleInstructionIfPossible
 When this method was called from visitAddition: in the Canonicaliser, it didn't take into account the fact that the addition of the two numbers may overflow. Now it does.


guillep:
Changes are good, they fix many broken tests :) Thanks @doste!

PR #131, opened 2024-03-22
More new primitives

PalumboN:
## Fixes

- Interface between fronted and trampolines: I changed the trampoline definitions in DruidJIT to use always the `TempReg` to send 1 parameter and get the result. Not implemented for >1 parameter (I think that I left a halt in that case). 
- Fix some cases of spilling (discovered by `primitiveAtPut` compilation).
- After the last changes in the canonicalizer, we lost some optimizations (like staging) because we are performing transformation in the CFG. So I have to **touch the list of optimisations** 😱  
- Small fixes / refactors...

## New primitives

- Many `primitiveFloat*` were supported without problem.
- `primitiveIdentityHash` -> after fixing the trampolines.
- `primitiveAs[Immediate]` -> add intrinsics in the interpreter

# Future work

### `primitiveValue` for _many args_
- Now Druid is compiling it only for `args = 0` version.
- We need to support meta-compilation of _variable arguments_.

### `primitive(Small)FloatDivide`
- Fix float comparison: `self CmpCq: 0.0 Rd: DPFPReg1` is not implemented.

### `primitiveStringReplace`
- It expects >2 args, so they travel through the stack (similar to the SimpleStackCogit calling convention).

### `primitiveAtPut`
- I was going to compile this primitive, fixing many bugs in the meta-compiler and bah.... BUT I arrived at a problem that is better to see in another issue.
- Now the primitive "works" but it is returning a trash value:

- The problem is a phi function created during tail duplication:
  - since we duplicate a block, then new merge points are created, and the phi should be propagated. See `DRBasicBlock >> mergeInstructionsWithDuplicatedBlock: duplicatedBlock withSuccessors: successors usingUsers: getUsersBlock`
  - this algorithm creates phis of phis based on the dominator tree (ok), and it leaves `Null` values if any value dominates the phi (bad).
  - in this case, the problem is on splitting the `exitBlock`, which has many predecessors.
  - this is a "real" problem because, in the end, a copy to Null is saved as _result_:  
image
  - but the _CodeGenerator_ is ignoring the copies to Null:
https://github.com/Alamvic/druid/blob/6f8878a4bec2a52d2f7f09171b494affc9bd3da5/Druid/DRCogitCodeGenerator.class.st#L851-L855
  -  So at the end, the Cogit code is copying trash to the `ReceiverResultReg` 👎


guillep (approved):
Good!

PR #132, opened 2024-03-28
Primitive value multiargs

PalumboN:
VM: https://github.com/pharo-project/pharo-vm/pull/763/

Support meta-compilation of primitives with variadic args (_multiargs_):

Using the pragma like this:
```st
<numberOfArguments: #(0 1)>
```

Druid will compile a generator with the pattern: 
```st
	s1 := self methodNumArgs.
	s1 = 0 ifTrue: [ ... ]

	s2 := self methodNumArgs.
	s2 = 1 ifTrue: [ ... ]

        ^ 0
```

And also leave the entry in the primitiveTable without arguments (`-1` at the end).

------

In this PR I also:

- Removed the `<argumentCountCheck>` customization -> we can have the same behaviour with these changes
- Fix some cases of Phi Functions with Null due to bad selection of the possible new operands under block splitting.
- Fix many tests, I left skipped the more complicated (I'll open issues for that).


PalumboN:
Some related tests are broken, I'll check them.

PalumboN:
Tests are fixed (or most of them). We need to merge VM: https://github.com/pharo-project/pharo-vm/pull/763 before.

Druid-Tests/DRDeadPathEliminationTest.class.st
```diff
@@ -85,7 +84,7 @@ DRDeadPathEliminationTest >> testEliminateDoubleDeadPaths [
 	| cfg lastBlocks lastPaths lastPathsLeft lastPathsRight |
 	cfg := self generateDruidIRFor: #primitiveDNA.
 
-	3 timesRepeat: [ 
+	4 timesRepeat: [
```
guillep, line 87: cuack :)

Druid/DRVMState.class.st
```diff
@@ -213,7 +213,7 @@ DRVMState >> simplify [
 { #category : #accumulating }
 DRVMState >> stackOperandAt: stackPosition [
 
-	^ operandStack at: stackPosition
+	^ operandStack at: stackPosition ifAbsent: [ DRNullValue new ]
```
guillep, line 216: This could be dangerous

guillep (approved):
Ok!

PR #151, opened 2024-05-02
Super bytecodes compilation

PalumboN:
Add the support for compiling _super instructions_.

- The meta-compiler receive a list of bytecodes to compile.
- It compiles all of them, in sequence, into one method.
- The first bytecode is annotated with the next one to combine
  - It checks the next bytecode at the beginning of the method, and executes the super instruction if applies
  - It only supports the combination of 2 bytecodes (no more yet)
  - It can have many possible _next_ bytecodes (with many guards)


PalumboN:
Ubuntu CI is passing 🟢

PR #156, opened 2024-07-04
Fix super-instructions for jumps

PalumboN:
Fix super-instructions to support bytecode jumps. There were two main problems:

1. Regarding the `annotateBytecode:`, there were 2 for the same super-bytecode, at the end. I added a pass `DRAnnotateBytecodeInstructionCoalescing` To move all of these instructions at the beginning. I'm not sure about this, I left some comments.

2. When another bytecode jumps to the middle of the super-instruction (to the 2nd bytecode). It is not managing correctly the `fixup`s. Now the superinstruction is avoided on those cases. I couldn't make it work as I wanted.

You can see the changes in `DruidTestRTLCompiler` to see the result of these fixes.


PalumboN:
Tests are green 💚 🙌

PalumboN:
Tests are green and I added a compilerOption to control super-instructions.

Merge is safe and we can move forward.

Issue #157, opened 2024-08-22
druidExitPoint does not work for meta-compilation of bytecodes


PalumboN:
I found that the definition for `pushNewArrayBytecode` generated by Druid does not manage a border case well.

This bytecode need to instantiate an array, so there is a check using the `objectMemory getScavengeThreshold` to check if the new object fits in the _new space_:
https://github.com/pharo-project/pharo-vm/blob/eaa8ca205aa46fe79bd87dae2cb75244f67e743e/smalltalksrc/VMMaker/DruidJIT.class.st#L12691-L12695

#### The problem is that, after the comparison, there is a jump to the end of the bytecode in case of "not enough space".

This happen because [the check in the interpreter](https://github.com/pharo-project/pharo-vm/blob/eaa8ca205aa46fe79bd87dae2cb75244f67e743e/smalltalksrc/VMMaker/SpurMemoryManager.class.st#L2516-L2535) uses `druidExitPoint` instrinsic.
```st
	freeStart + numBytes > scavengeThreshold ifTrue: [
		self druidExitPoint.
		needGCFlag ifFalse: [self scheduleScavenge].
		self validateRoomInEdenOrDie: numBytes ].
```

But that intrinsic is implemented to **work only with primitives**:
https://github.com/Alamvic/druid/blob/742b6c936f8f5d09c4052c3f310805e124be1e46/Druid/DRMetaCompilerIRGenerator.class.st#L62-L78

That is why it is jumping at the end of the definition (for primitives, it is the way to "fail" and execute the fallback).

### We need to merge `druidExitPoint` (for primitives) with `druidForceIntepretation` (for bytecodes).

PR #161, opened 2024-12-03
Fix vm

PalumboN:
Fixing many bugs found running the generated VMs.

- Fix special register allocation for closure creation
- Fix https://github.com/Alamvic/druid/issues/157 (we should keep only one)
- Fix generation of `annotateBytecode:`
  - Now this logic is only during code generation (no instruction in the CFG), I am not sure about what's better.
    - It adds it after any `isMapped` trampoline
    - Of if the bytecode is annotated (in the table) there is a check when the generator arrives to the return (for the cases of staged branches)
- There where some problems about _IMMUTABILITY_  and _write barrier_ checks
  - The meta-compilation was generating 2 checks (one for Old Space and other for Perm Space), but both call to the same trampoline. For now I ignored permi from Druid, but we should re-write the interpreter to support both but only call the trampoline once (or maye not.. I'm not sure).
  - I force a `IMMUTABILITY := true` during code generation, because the code is different and it is a global variable that can have any value.
  - I also changed the order of pops and deoptimization points in the interpreter for better code in Druid

---

We need to merge the PR in the VM before this https://github.com/pharo-project/pharo-vm/pull/880


guillep:
Loading?

PalumboN:
Yes, github was slow 👀  Now it is ok.

PalumboN:
I'll merge this, I'm working on the next.

PR #164, opened 2025-02-24
Improve code generation

PalumboN:
Fix:

- `nilObject` type
- Reuse registers between stack dependencies
- Refactoring / cleaning code


PalumboN:
The two failing tests are identified and they are related with test code in the VM. I'll merge and fix it in the next PR.

PR #165, opened 2025-02-28
Improve dropal 12 - Support compiling blocks

PalumboN:
Now we can compile methods with blocks, and also with complex CFG (as `select:`)

### What's done:

- Change the generator algorithm to avoid "tail duplications" by visiting the same block many times (now it iterates in reverse post order).
- Change the previous last pass to make the CFG "byteable" -> first, remove all Phis (new); then reuse values (old).
- Some changes around types that could make interference with DrCogit stuff.
- Pass on tests

### What's missing:

- Compile blocks accessing method arguments


Ducasse:
Oh this is cool. A nice important step.

PR #166, opened 2025-03-13
Update pharo 12

PalumboN:
- Druid now runs on Pharo 12 images
- It works with the VM on pharo-12 also: https://github.com/pharo-project/pharo-vm/pull/929
- Not using PCG anymore (check if something is missing and remove it from the baseline!)


PalumboN:
Tests are green 🟢 ([for linux](https://github.com/Alamvic/druid/issues/170))
Merge

guillep:
🙌

PR #173, opened 2025-04-10
Collapse bit and instructions with comparisons

m-demare:
Convert `TestEqualsThan(BitAnd(a, b), 0)` into `TestNoMask(a, b)`
and `TestNotEqualsThan(BitAnd(a, b), 0)` into  `TestMask(a, b)`

Add codegen for `DRNoTestMask`

Add codegen tests for `DRTestMask` and `DRTestNoMask`, to validate thay don't emit a Cmp


Druid/DRCPSEdge.class.st
```diff
@@ -580,6 +580,15 @@ DRCPSEdge >> visitNoOverflow: aDROverflow inBranch: aDRBranch [
...
 ]
 
+{ #category : 'visiting' }
+DRCPSEdge >> visitNoTestMask: aDRTestMask inBranch: aDRBranch [
+
+	^ self
+		  visitComparison: aDRTestMask
+		  ofBranch: aDRBranch
+		  andDo: [ :reg :value | (DRRegisterDomain mask: reg with: value) negated ]
+]
```
PalumboN, line 590: 😉

Druid/DRCogitCanonicaliser.class.st
```diff
@@ -24,6 +24,42 @@ DRCogitCanonicaliser >> applyTo: cfg [
...
+
+	| oldCondition newCondition newInstruction |
+	oldCondition := aDRGetConditionCode condition.
+	(oldCondition isEqualsThanComparison or:
+		 oldCondition isNotEqualsThanComparison) ifFalse: [ ^ aDRGetConditionCode ].
+
+	"Move constant, if present, to the right"
+	aDRGetConditionCode operand1 isConstant ifTrue: [
+		aDRGetConditionCode operands:
+			{ oldCondition }, aDRGetConditionCode operands reverse ].
```
PalumboN, line 40: We should `reverse` the condition too. I understand that here is the same, because we are checking by _equality comparison_ and it is commutative.
I think that it's better to be explicit:
```suggestion
	"Move constant, if present, to the right"
	aDRGetConditionCode operand1 isConstant ifTrue: [
		aDRGetConditionCode operands:
			{ oldCondition reversed }, aDRGetConditionCode operands reverse ].
```

Druid/DRCogitCodeGenerator.class.st
```diff
@@ -1189,6 +1189,14 @@ DRCogitCodeGenerator >> visitNoOverflow: aDRNoOverflow inBranch: aDRBranchIfCond
 	^ aDRBranchIfCondition 
 ]
 
+{ #category : 'visiting' }
+DRCogitCodeGenerator >> visitNoTestMask: aDRNoTestMask inBranch: aDRBranchIfCondition [
+	
+	self visitTestMask: nil inBranch: aDRBranchIfCondition.
+	
+	^ aDRBranchIfCondition reversed
+]
```
PalumboN, line 1198: Why a `nil` here?

Druid/DRNoTestMask.class.st
```diff
@@ -6,8 +6,32 @@ Class {
...
+DRNoTestMask >> opcode [
+	
+	^ 'NoTest'
+]
+
+{ #category : 'converting' }
+DRNoTestMask >> reversed [
+	"Returns the same condition but with reversed operands: NoTest x y --> NoTest y x"
+	^ DRNoTestMask new.
+]
```
PalumboN, line 37: Yes, commutative constraints return always `self`:
```suggestion
DRNoTestMask >> reversed [
	^ self
]
```

PalumboN (dismissed):
Niceee!!! 🚀  

I left some comments but they are not critical.

Druid/DRCogitCanonicaliser.class.st
m-demare: Yeah, makes sense

Druid/DRCogitCodeGenerator.class.st
m-demare: Because that method's parameter is a `DRTestMask`, and I don't have one here (I have a `DRNoTestMask`). I thought of passing that, but I think it'd be more error-prone in the future, since someone may assume it's always a `DRTestMask`
That argument isn't used anyways, so it doesn't make much of a difference

PalumboN: Ookkkkk maybe would be nice to have a comment :)

Issue #174, opened 2025-04-10
Compile instVar assigments


PalumboN:
Now, after inlinings, we probably want to update an _instance variable_ from another object different to `self`.

So, we are compiling them as message sends of `instVarAt:put:`.

BUT, when this message is sent to `self`, we can compile a normal assignment.

For example:

Now we are generating this

```st
exampleNotNilIfNotNilWithArgument

	^ self instVarAt: 1 put: 1
```
That could be written as
```st
exampleNotNilIfNotNilWithArgument

	^ iVar := 1
```

PR #178, opened 2025-04-22
Optimization InstVar 

luc-raz:
when accessing or assigning variable to self, use instead of instVarAt: and instVarAt: put:

Fix: #174


Druid-Tests/DruidTestInterpreter.class.st
```diff
@@ -24,6 +27,41 @@ DruidTestInterpreter >> basicInlineMethod [
...
+{ #category : 'as yet unclassified' }
+DruidTestInterpreter >> basicInlineMethodAccesInstVar [
+	<var: #aStack type: #Stack >
+	
+	| aStack |
+	aStack := Stack new.
+	aStack push: 'element'.
+	
+	^ aStack firstLink 
+]
```
PalumboN, line 39: Interesting 🚀

Druid/DRBytecodeGenerator.class.st
```diff
@@ -197,9 +197,16 @@ DRBytecodeGenerator >> visitLoadArgument: aDRLoadArgument [
 { #category : 'ir-to-target' }
 DRBytecodeGenerator >> visitLoadInstVar: aDRLoadInstVar [ 
 	
-	aDRLoadInstVar operand1 acceptVisitor: self.
-	builder pushLiteral: aDRLoadInstVar index.
-	builder send: #instVarAt:
+	aDRLoadInstVar operand1 acceptVisitor: self.
```
PalumboN, line 200: This should not be inside the if?

This visit push the object to access, but this is only necessary for the _message send case_ `send: #instVarAt:`.
The `pushInstVar` bytecode is always on the receiver (that's the original problem), so we don't need to push it before (it's implicit).

Druid/DRBytecodeGenerator.class.st
```diff
@@ -197,9 +197,16 @@ DRBytecodeGenerator >> visitLoadArgument: aDRLoadArgument [
 { #category : 'ir-to-target' }
 DRBytecodeGenerator >> visitLoadInstVar: aDRLoadInstVar [ 
 	
-	aDRLoadInstVar operand1 acceptVisitor: self.
-	builder pushLiteral: aDRLoadInstVar index.
-	builder send: #instVarAt:
+	aDRLoadInstVar operand1 acceptVisitor: self. 
+	
+	aDRLoadInstVar operand1 value class = DRLoadTemporaryVariable ifTrue: [
```
PalumboN, line 202: Mmmmm this condition is not enough, the object doesn't come _always_ from a variable (unless I'm missing some transformation).

Soooo.... We need a test to make this fail! 🔥 
Maybe something like:
```st
m
  ^ ('a' -> 'b') key
```

Druid/DRBytecodeGenerator.class.st
```diff
@@ -251,11 +258,20 @@ DRBytecodeGenerator >> visitReturn: aDRReturn [
 
 { #category : 'ir-to-target' }
 DRBytecodeGenerator >> visitStoreInstVar: aDRStoreInstanceVariable [ 
-
+	
 	aDRStoreInstanceVariable operand1 acceptVisitor: self.
```
PalumboN, line 262: Here also

Druid/DRBytecodeGenerator.class.st
```diff
@@ -251,11 +258,20 @@ DRBytecodeGenerator >> visitReturn: aDRReturn [
...
-	builder pushLiteral: aDRStoreInstanceVariable operand2 value.
-	aDRStoreInstanceVariable operand3 acceptVisitor: self.
-	builder send: #instVarAt:put:
+	aDRStoreInstanceVariable operand1 class = DRLoadTemporaryVariable ifTrue: [
```
PalumboN, line 263: And this condition is **repeated code**, what is something that we want to avoid.

The problem is a **missing abstraction**. I think we want a method 
```st
self isSelf: aDRStoreInstanceVariable operand1
```

Druid/DRBytecodeGenerator.class.st
```diff
@@ -251,11 +258,20 @@ DRBytecodeGenerator >> visitReturn: aDRReturn [
...
+	
 	aDRStoreInstanceVariable operand1 acceptVisitor: self.
-	builder pushLiteral: aDRStoreInstanceVariable operand2 value.
-	aDRStoreInstanceVariable operand3 acceptVisitor: self.
-	builder send: #instVarAt:put:
+	aDRStoreInstanceVariable operand1 class = DRLoadTemporaryVariable ifTrue: [ 
+		builder pushLiteral: aDRStoreInstanceVariable operand2 value.
+		aDRStoreInstanceVariable operand3 acceptVisitor: self.
+		builder send: #instVarAt:put: .
+		builder popTop.
```
PalumboN, line 267: Niceeeee 💯

Druid/DRBytecodeGenerator.class.st
```diff
@@ -251,11 +258,20 @@ DRBytecodeGenerator >> visitReturn: aDRReturn [
...
+	]
+	ifFalse: [ 
+		aDRStoreInstanceVariable operand3 acceptVisitor: self.
+		builder storeInstVar: aDRStoreInstanceVariable operand2 value.
+		builder popTop.
+	].
```
PalumboN, line 273: I suggest to write these methods in the form:

```st
CONDITION ifTrue: [
   ... TRUE CODE ...
  ^ self
]
... FALSE CODE ...
```

_Generally_, a simple ifTrue: is easier to read than an `ifTrue:ifFalse:`

PalumboN (commented):
Good job! 👏 👏 👏 

I left some comments to improve the code. And also we need more tests, because (I think) the current implementation does not work for every case.

Druid/DRConstantValue.class.st
```diff
@@ -143,6 +143,7 @@ DRConstantValue >> type [
 	(#( true false ) includes: value) ifTrue: [ ^ DRBooleanType new ].
 	value isInteger ifTrue: [ ^ DRSignedIntegerType size: 8 "Word size" ].
 	value isFloat ifTrue: [ ^ DRFloatType new ].
+	value isGlobalVariable ifTrue: [ ^ value value class asDRType ].
```
PalumboN, line 146: And we should check this condition, some tests are failing with MNU:
```
 ✗ #testCompareRefenceValueFalse (34ms)
Instance of Spur64BitMMLECoSimulator did not understand #isGlobalVariable
Spur64BitMMLECoSimulator(Object)>>doesNotUnderstand: #isGlobalVariable
DRConstantValue>>type
DRPrimitiveIRGenerator(DRIRGenerator)>>methodFor:withReceiver:
DRPrimitiveIRGenerator(DRMetaCompilerIRGenerator)>>methodFor:withReceiver:
DRPrimitiveIRGenerator(DRMetaCompilerIRGenerator)>>resolveMessageSend:receiver:arguments:
DRPrimitiveIRGenerator(DRIRGenerator)>>resolveMessageSend:receiver:
DRPrimitiveIRGenerator(DRIRGenerator)>>visitMessageNode:
RBMessageNode>>acceptVisitor:
DRPrimitiveIRGenerator(DRIRGenerator)>>visitOperand:
DRPrimitiveIRGenerator(DRIRGenerator)>>interpretCondition:conditionKind:
DRPrimitiveIRGenerator(DRIRGenerator)>>interpretGreaterThanComparisonWith:
```

Druid/DRBytecodeGenerator.class.st
luc-raz: If we do this ('a' -> 'b'), the inlining will not happen (it will never enter this method), because it will not know the type. The pragma need the name of the variable.

Druid/DRBytecodeGenerator.class.st
luc-raz: Same problem as above

PalumboN: It's not the same problem, here we are pushing an extra value to the stack when the object is self. This happens also without inlinings.

Druid/DRBytecodeGenerator.class.st
PalumboN: Mmmmm because our typesystem is _shitty_. We need to improve it the next days.

With numbers neither? `1 -> 2`?

Druid/DRBytecodeGenerator.class.st
luc-raz: Already fixed

Druid/DRBytecodeGenerator.class.st
luc-raz: We can re-test it if after we implement new type sytem

Druid/DRBytecodeGenerator.class.st
```diff
@@ -250,12 +261,19 @@ DRBytecodeGenerator >> visitReturn: aDRReturn [
 ]
 
 { #category : 'ir-to-target' }
-DRBytecodeGenerator >> visitStoreInstVar: aDRStoreInstanceVariable [ 
-
+DRBytecodeGenerator >> visitStoreInstVar: aDRStoreInstanceVariable [
+
+	(self isSelf: aDRStoreInstanceVariable) ifTrue: [
```
PalumboN, line 266: Here I would expect `self isSelf: aDRStoreInstanceVariable operand1` (the store is not self, neither the load). It's more clear.

Also, it's coincidence that both (load and store) have the receiver object as `operand1`. It has no too much sense to have it in `isSelf:`.

PalumboN (approved):
This is super cool! 🚀 💯

Issue #169, opened 2025-03-19
Avoid meta-compilation on each test


PalumboN:
Now each test is meta-compiling the primitive/bytecode under assertion.
This causes the same primitive/bytecode is meta-compiled multiple times.

**It is too time consuming.**

We should meta-compile all primitive/bytecodes for Register/Simple Cogit once at the beginning, and the tests should only execute the already generated code on the machine code simulator and do the asserts.

_Check the `TestResource` class_

PR #189, opened 2025-05-12
Run bytecode meta-compilation tests fast

PalumboN:
Fix https://github.com/Alamvic/druid/issues/169

Before
image

After
image

------

This PR introduce the `DRBytecodeCompilationTestResource` resource for the tests compiling bytecode and primitives (just Production).

The idea is:
- Do all the meta-compilations once on the resource
- Each test just invoke the generator and test the compiled code on Unicorn

I also removed all dependencies to PharoCodeGenerator (dead code)

-------

After the refactor, I would just say that the problem were the `DRProductionPrimitiveCompilationTest`. 
We didn't gain so much time from bytecode tests...
So I didn't do it for the "ScenarioPrimitives", because this workflow is more complex and it's not worth if the time doesn't reduce significantly.


Druid-Tests/DRProductionBytecodeTest.class.st
```diff
@@ -1276,29 +994,22 @@ DRProductionBytecodeTest >> testBytecodePushReceiverDoesNotOverridePreviousPush
...
+		"Push receiver, then pop it"
+		self compileBytecodeSelector: #pushReceiverBytecode.
+		cogit ssPopTopToReg: (cogit allocateRegNotConflictingWith: 0).
+		
+		cogit genReturnTopFromMethod ].
 
 	self
-		executeMethodWithReceiver: (memory integerObjectOf: 17)
+		executeMethodWithReceiver: memory nilObject
 		withArguments: #( 42 ).
```
PalumboN, line 1013: This test is failing (just) on Ubuntu, now 🤔

Druid-Tests/TestAsserter.extension.st
```diff
@@ -0,0 +1,7 @@
+Extension { #name : 'TestAsserter' }
+
+{ #category : '*Druid-Tests' }
+TestAsserter >> onPharoCITestingEnvironment [
+
+	^ Smalltalk os environment at: 'PHARO_CI_TESTING_ENVIRONMENT' ifPresent: [ true ] ifAbsent: [ false ]
+]
```
PalumboN, line 7: I moved this from the subclass, but I must add it as an extension until making a PR to Pharo.

PR #195, opened 2025-06-10
IR printer and parser

m-demare:
Implement a basic IR printer and parser. This seems useful for comparing CFG's, debugging CFG's, implementing tests, and even deep-copying CFG's

The representation is pretty self-explainatory, but here's the summary:

- It always starts with `[[initialBasicBlock]]`, which declares the name of the initial basic block. This block has to be later defined somewhere 
- Blocks have the syntax:
```
blockName:
    instruction list...
```
- Instructions have the syntax: `result := opcode(operands)`. (the `opcode` method was already defined, which caused some weird syntax things, like the opcode for Copy being `:=`)
- Instructions that don't care about the result, can use `_`
- Blocks can optionally specify the order of its predecessors, using `blockName:[block1, block]` at its definition. This is needed for blocks containing phi functions

See `testParseCyclicCFG` for a decent example


Druid-Tests/DRIRPrinterTest.class.st
```diff
@@ -0,0 +1,75 @@
...
+	cfg := self setUpCFGWithConditionalWithPhi.
+
+	instructionFactory := cfg instructionFactory.
+	condition := instructionFactory less: 1 than: 2.
+	cfg b1 addInstructionFirst: condition.
+	cfg b1 lastInstruction operands: {
+			DREqualsThanComparison new.
+			condition.
+			true asDRValue }.
+	cfg validate.
```
PalumboN, line 22: If we want to check this, we should do it at the end before printing (I think). 
We can have a method `self print: cfg` to check and use the printer ;)

Druid-Tests/DRIRPrinterTest.class.st
PalumboN, line 1: No `testPrintCyclicCFG` here? :P

Druid/DRBranchIfCondition.class.st
```diff
@@ -12,6 +12,25 @@ Class {
...
+	operands := self processDependencyOperands: operands.
+
+	conditionClasses := DRCondition allSubclasses.
+	condition := (conditionClasses at:
+		              (conditionClasses findFirst: [ :cl |
+			               cl new opcode = operands first ])) new.
+	operands at: 1 put: condition.
+
+	^ { operands. targets }
+]
```
PalumboN, line 32: ⚠️  I'm not sure to have this behaviour in the class, I was expecting to have it on the Parser

Druid/DRBranchIfCondition.class.st
```diff
@@ -207,6 +226,13 @@ DRBranchIfCondition >> prettyPrint [
 	  , (self falseBranch ifNotNil: [ :br | br id ]) asString
 ]
 
+{ #category : 'printing' }
+DRBranchIfCondition >> printIROperands [
+
+	^ condition opcode , ', ' , super printIROperands , '; '
+	  , (', ' join: (self targets collect: [ :b | 'b' , b id asString ]))
+]
```
PalumboN, line 234: ⚠️ Idem this, I would expected to have it in the Printer

Druid/DRNoop.class.st
```diff
@@ -41,6 +41,11 @@ DRNoop >> removeFromCFG [
 	super removeFromCFG
 ]
 
+{ #category : 'accessing' }
+DRNoop >> result: _ [
+
+]
+
```
PalumboN, line 48: Why this??

Druid/DRMethod.class.st
```diff
@@ -9,6 +9,12 @@ Class {
 	#tag : 'CompilerCompiler'
 }
 
+{ #category : 'testing' }
+DRMethod class >> acceptsOpcode: anOpcode [
+
+	^false
+]
```
PalumboN, line 16: This should not be the default in some superclass? 

Because now it's just defined in some of the DRInstruction

PalumboN (commented):
Cooooolllll 😎 

I leave some comments...

The **most important** one(s) are the related to having the _parsing_ and _printing_ responsibilities in the IR classes. For me that should be implemented outside of the hierarchy, inside the parser/printer objects, probably as a frontend/backend extension (implemented as visitors).

_Think that we have many front/mid/backends that do different things with the IR. If we put all the code from all of them in the IR, we finish with god objects in the nucleus of the system. Difficult to maintain and extend, because we need to update all the things on any change at the same moment. That is why we have a design where each frontend/backend implement the logic outside of the IR hierarchy (mainly as visitor pattern) in different packages. And we just leave in the IR the "pure things" (that we decide at each moment of the development)_

Druid-Tests/DRIRPrinterTest.class.st
m-demare: oops was left from debugging, doesn't make much sense there tbh

Druid-Tests/DRIRParserTest.class.st
```diff
@@ -0,0 +1,122 @@
...
+	cfg := parser parse.
+	printer := DRIRPrinter for: cfg.
+
+	self assert: printer asString equals: '[[b0]]
+
+b1:
+	t0 := LessThan(1, 2)
+	t1 := :=(5)
+	t2 := MUL(3, t1)
+	t3 := If(EqualsThan, t0, true; b2, b3)
```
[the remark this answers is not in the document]
m-demare: >  ⚠️ Ahh and also the If should not have result (in general the last block's instruction)

This is fixed now

PR #210, opened 2025-09-19
Improvement DRAbstractInstruction>>validate

m-demare:
Check `self isDominatedBy: o` instead of `self basicBlock isDominatedBy: o basicBlock`. This way:

- If they're in the same block, it also checks if the operand comes before the usage
- It can also handle phi functions (I think, I guess CI will tell us for sure)

Merge after https://github.com/Alamvic/druid/pull/209

### DIFF
--- a/Druid/DRAbstractInstruction.class.st
+++ b/Druid/DRAbstractInstruction.class.st
@@ -232,11 +232,11 @@ DRAbstractInstruction >> validate [
 	self validateDependenciesAreUsedByMyself.
 	self validateUsersUseMyself.
 
-	self isPhiFunction ifTrue: [ ^ self ].
 	self operands do: [ :o |
-		((o isInstruction not or: [
-			  self basicBlock isDominatedBy: o basicBlock ]) or: [ "this is to handle the case we have different CFG, block is interpreted with a new CFG"
-			 self controlFlowGraph ~= o controlFlowGraph ]) ifFalse: [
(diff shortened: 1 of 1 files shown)


Druid/DRAbstractInstruction.class.st
```diff
@@ -232,11 +232,11 @@ DRAbstractInstruction >> validate [
...
 	self validateUsersUseMyself.
 
-	self isPhiFunction ifTrue: [ ^ self ].
 	self operands do: [ :o |
-		((o isInstruction not or: [
-			  self basicBlock isDominatedBy: o basicBlock ]) or: [ "this is to handle the case we have different CFG, block is interpreted with a new CFG"
-			 self controlFlowGraph ~= o controlFlowGraph ]) ifFalse: [
+		(o isInstruction not or: [
+			 (self isDominatedBy: o) or: [
+				 "this is to handle the case we have different CFG, block is interpreted with a new CFG"
```
guillep, line 238: The comment is old not valid anymore no?

guillep: Or is it about the code below?

When can we have instructions using operands from a different cfg? This looks broken semantics.
@PalumboN ideas?

PalumboN: I have no idea, but for me each CFG is independent from others

PR #216, opened 2025-09-29
Improve optimization order, so that they apply faster in big CFG's

m-demare:
Adding more of the low-cost/high-reward optimizations at the beggining, simplifying the CFG for the latter, more expensive optimizations, greatly reduces total compile time
With this change, #primitiveAtPut compiles about 1.5 min faster


PalumboN:
This PR makes the metacompilation process faster, but **produces less quality code**.

I pushed some results of metacompilation to analyze. In a quick check I identified:
- Not using `TstCq:R:`
- Spilling more values to memory
- Some dead path not eliminated (maybe?)

_This needs more work yet_

PalumboN (changes requested):
See my comment ⬆️

PR #218, opened 2025-10-17
Compile Opal examples

PalumboN:
I was doing a pass on DrOpal:
- compiling more Opal examples (there are some missing yet)
- adding tests
- fixing buggy cases

### The most importan change here is the use of scopes (particularly when inline)

**Before** we were renaming the temporaries to avoid collisions after method inlining.

**Now** we are creating a scope for each inlining (so a method's scope could have many "inlined" scopes).
And each scope generates a _temp vector_ (DRInitializeScope).

**With that**:
- there are no collisions (temps are accessed from different vectors)
- It is not necessary to do the rename (and manage that)

### DIFF
--- a/Druid-Opal/DRMethodIRGenerator.class.st
+++ b/Druid-Opal/DRMethodIRGenerator.class.st
@@ -61,7 +61,6 @@ DRMethodIRGenerator >> initializeArgumentsNames: aDRCode with: argumentsValues [
 
 { #category : 'initialization' }
 DRMethodIRGenerator >> initializeIR [
-	"Initialize scope temps"
 
 	super initializeIR.
 	self currentBasicBlock initializeScope: self scope
@@ -123,7 +122,7 @@ DRMethodIRGenerator >> interpretAssignmentNode: aRBAssignmentNode [
 				          addInstructionWithNoResultFrom: aRBAssignmentNode
 				          instructionKind: DRStoreTemporaryVariable
 				          operands: {
-						          aRBAssignmentNode variable name asDRValue. "Fix temporary names collisions"
+						          aRBAssignmentNode variable name asDRValue.
 						          value }.
 
 			result scope:


Druid-Opal/DRBlockIRGenerator.class.st
```diff
@@ -44,10 +38,30 @@ DRBlockIRGenerator >> variableFrame [
...
 { #category : 'accessing' }
-DRBlockIRGenerator >> variableFrame: aDRStackFrame [ 
+DRBlockIRGenerator >> variableFrame: aDRStackFrame [
+
 	variableFrame := aDRStackFrame
 ]
 
+{ #category : 'accessing' }
+DRBlockIRGenerator >> visitArgumentVariableNode: aRBVariableNode [
+	"TODO: Repeat code"
```
PalumboN: TODO

Druid-Opal/DRMethodIRGenerator.class.st
```diff
@@ -130,20 +154,83 @@ DRMethodIRGenerator >> interpretCode: aDRMethod receiver: aReceiver [
...
+	"We add new temporaries to the current scope"
+
+	self scope addTempVariables:
+		(aRBBlockNode temporaries collect: [ :temp |
+			 temp variable originalVar ]).
+	self scope addTempVariables:
+		(aRBBlockNode arguments collect: [ :temp | temp variable ]).
+
+	self interpretBlockBody: aRBBlockNode
+]
```
PalumboN: TODO: Create an inline scope?

Add tests with collisions?
```
[ |x| ... ] whileTrue: [ |x| ... ]
```

Druid-Opal/DRMethodIRGenerator.class.st
```diff
@@ -260,26 +345,22 @@ DRMethodIRGenerator >> visitArgumentVariableNode: aRBVariableNode [
...
-			^ self
-				  addInstructionFrom: value
-				  instructionKind: DRCopy
-				  operands: { value operand1 } ].
-		^ self pushOperand: value ].
+		value := self temporaryAt: varName.
+
+		"TODO: Avoid Nulls here (do not initialize)"
+		value simpleConstantFold asDRValue isNullValue ifFalse: [
+			^ self pushOperand: value ] ].
```
PalumboN: TODO. But I'm not really sure about that

Druid-Opal/DRUnconditionalBackJumpScheluder.class.st
PalumboN: TODO. Add tests

Druid/DRMessageSend.class.st
PalumboN: We need to re-check this. Now the blocks knows the IR, so we do not need to talk with the `inlineGenerator` here. Maybe we don't need the `innerDependencies` anymore?

PR #245, opened 2026-04-10
Fix multi inline

### LINKED ISSUE #224: Fix multiple inlinings scopes
We need to create a new scope for each inlined method, but this case is managed by the same IR generator (so the same scope...)

_Originally posted by @PalumboN in https://github.com/Alamvic/druid/pull/218#discussion_r2478442092_

PalumboN:
Fix https://github.com/Alamvic/druid/issues/224

 scopes creation for the cases of multiple methods inlining on the same message send

Before, all inlined methods were sharing the same scope

Now, each one creates its own scope. They have as outer the same empty scope (from the send) that connects with the caller method


luc-raz:
All good for me
Ready to merge

PR #249, opened 2026-04-22
Fix inlined blocks

### LINKED ISSUE #242: Problems compiling inlined blocks
Try to compile this method and fails to compile: `hallo := 1`

```st
exampleSimpleBlockLocalWhile
	|a|
	a := true.
	^[: b | [a] whileTrue: [ | hallo |  a := false. hallo := 1 . hallo]] value: 1
```

luc-raz:
Fix: #242


Druid-Opal/DRBytecodeGeneratorTest.class.st
```diff
@@ -301,6 +313,12 @@ DRBytecodeGeneratorTest >> testMethodReadingClassVariable [
 	self assertCompilationFor: #methodReadingClassVariable fromClass: DrOpalExamples
 ]
 
+{ #category : 'tests' }
+DRBytecodeGeneratorTest >> testMethodSimpleBlockLocalWhile [
+
+	self assertCompilationFor: #exampleWhileWithTemp fromClass: OCOpalExamples
```
PalumboN, line 314: I started to move the example methods to a class in this project because they can change when we update the Pharo version

Druid/DRExecutionStack.class.st
```diff
@@ -41,7 +41,7 @@ DRExecutionStack >> baseFrame [
 { #category : 'enumerating' }
 DRExecutionStack >> detectFrame: aFullBlockClosure [ 
 	
-	^ stateStack detect: aFullBlockClosure
+	^ stateStack detect: aFullBlockClosure ifNone: [ ^ nil ]
```
PalumboN, line 44: Mmmmm I'm not sure about this change. If `detectFrame:` returns nil we will have problems later...

Druid/DRClosureStackFrame.class.st
```diff
@@ -10,9 +10,10 @@ Class {
 DRClosureStackFrame >> frameDefining: aName withState: aState [
 
 	^ (super frameDefining: aName withState: aState) ifNil: [
-		  (self outerFrameWithState: aState)
-			  frameDefining: aName
-			  withState: aState ]
+		  (self outerFrameWithState: aState) ifNotNil: [ :frame | 
+			frame frameDefining: aName withState: aState ] 
+		ifNil: [ ^ nil ]
```
PalumboN, line 15: Idem here

Druid-Opal/DRMethodIRGenerator.class.st
```diff
@@ -167,14 +167,45 @@ DRMethodIRGenerator >> interpretCode: aDRCode receiver: aReceiver arguments: arg
...
 DRMethodIRGenerator >> interpretInlinedBlockBody: aRBBlockNode [
 	"We add new temporaries to the current scope"
-
-	self scope addTempVariables:
+	| currentScope newScope|
+	
+	currentScope := self scope.
+	
+	aRBBlockNode scope tempVector ifNotEmpty: [ | tempVectorVar |
+		"case where in the inlined block, a tempVector should be created"
```
PalumboN: So the problem is that a scope was missing? I see that before it was adding the variables to the outerScope (the method?). Why that was not working?

Druid-Opal/DRMethodIRGenerator.class.st
```diff
@@ -167,14 +167,45 @@ DRMethodIRGenerator >> interpretCode: aDRCode receiver: aReceiver arguments: arg
...
+		currentScope resolveScopeId. "shift each id by one"
+		
+		tempVectorVar := aRBBlockNode scope tempVector values.
+		newScope addTempVariables: tempVectorVar .
+		
+		newScope addTempVariables: 
+			((aRBBlockNode temporaries collect: [ :temp |
+			 	temp variable originalVar ]) difference: tempVectorVar) .
+		
+		self currentBasicBlock initializeScope: newScope.
```
PalumboN: We already have this code to initialize the scopes, we should reuse it. I think that we want a InlineIRGenerator (it already creates it's own scope).

Druid-Opal/DRMethodIRGenerator.class.st
```diff
@@ -167,14 +167,45 @@ DRMethodIRGenerator >> interpretCode: aDRCode receiver: aReceiver arguments: arg
...
+		self scope: newScope .
+		self interpretBlockBody: aRBBlockNode.
+		
+		self scope: currentScope.
```
PalumboN: Mmmmmm I prefer to avoid this kind of things, could be a hell to debug later.

I think that with my idea of the inlineIRGenerator it is not necessary 🤔

PalumboN (commented):
Nice @luc-raz 🚀 

I put some comments, I'd say to have a meeting next week and talk about them

Druid/DRExecutionStack.class.st
PalumboN: Let's add a comment about the case where the frame is `nil`. What that means?

"That means that we are in an inlined block (whileTrue: ifTrue:) accessing to a variable defined inside the block"

Druid-Opal/DRMethodIRGenerator.class.st
PalumboN: Check the method `interpretInlinedCode: code receiver: receiver arguments: arguments`

This already:
- Creates an inline generator (with its own scope)
- Jump to a basic block for generating the code (by the inline)
- Interpret the code
- And then restore the state to the original generator

With that, we don't need to manage the scopes at this level

PR #252, opened 2026-05-15
Fix inlining: super + nil + non-local returns

### LINKED ISSUE #247: Inline + super = bug
We are not inlining super sends correctly. The type system retrieves the lookup for `self` instead of the super class.

PalumboN:
- Fix inlining of super sends Fix https://github.com/Alamvic/druid/issues/247
- Changes in Null vs nil (cogit vs opal)
- Clean code
- Adding an assertion to fail compilation in case of "unresolved" non-local returns (the method was inlined but not the block). With that is easier to setup the benchmarks that I'm trying to do. We can iterate on the next steps.

~~⚠️  Something to check is the impact of this in the Cogit generated code 🧐~~


Druid-Opal/DRBytecodeBlockGenerator.class.st
```diff
@@ -31,7 +40,23 @@ DRBytecodeBlockGenerator >> generateTargetASTFromIR: aDRControlFlowGraph [
...
+
+	aDBlockRReturn operand1 acceptVisitor: self.
+	builder blockReturnTop.
+]
+
+{ #category : 'ir-to-target' }
+DRBytecodeBlockGenerator >> visitReturn: aDRReturn [ 
+
+	blockScope isInlined ifTrue: [ 
+		Error signal: 'Non-local return on inlined method breaks the semantics' ].
```
luc-raz, line 59: is this only for inlinedBlock (whileTrue, ifTrue) or all the block?

Druid-Opal/DRCleanScopes.class.st
```diff
@@ -9,29 +9,59 @@ Class {
...
+
+{ #category : 'as yet unclassified' }
+DRCleanScopes >> cleanTemporary: temp from: aDRScope withInstructions: allInstructions [
+
+	| variableInstructions |
+	variableInstructions := allInstructions select: [ :i | i variableName = temp ].
+	(self isUsedVariable: variableInstructions) ifFalse: [
+			variableInstructions do: [ :i |
+				i hasUsers ifFalse: [ i removeFromCFG ] ] ]
+]
```
luc-raz, line 55: this remove the instructions if not used. We don't have other optimisations that did the same operations?

Druid-Opal/DrOpalExamples.class.st
```diff
@@ -97,6 +97,12 @@ DrOpalExamples >> blockWithValue: arg [
 	^ [ arg ]
 ]
 
+{ #category : 'displaying' }
+DrOpalExamples >> displayString [ 
+
+	^ super displayString , ' cool'
```
luc-raz, line 103: What is cool

Druid-Opal/DRBytecodeBlockGenerator.class.st
PalumboN: For all is checking if we are compiling a block that have a non-local return and the method attached to that return was inlined, that means that the return must jump to the middle of the method (we cannot do that, yet)

Druid-Opal/DRCleanScopes.class.st
PalumboN: We have, for example, dead code elimination. The problem here is that the SSA is broken under temporaries variables because of blocks... maybe we should re-think it to see if we can do it better

Druid-Opal/DrOpalExamples.class.st
PalumboN: We are cool 😎

PalumboN: This example is used in a test because the inline was not inlining the super definition. So, I needed a method that exists in Object class too

Druid-Opal/DRBytecodeBlockGenerator.class.st
luc-raz: the name of the function is not explicit, when we read it seems to check for all inlined block
Or it checks for specific situation

PalumboN: This method is in the DRBytecodeBlockGenerator, so we are already generating a block: it was not inlined.

It's true that the question: `blockScope isInlined` is not right. It should be `methodScope isInlined`

Druid-Opal/DRCleanScopes.class.st
PalumboN: https://github.com/Alamvic/druid/issues/253

Druid-Opal/DRBytecodeBlockGenerator.class.st
PalumboN: I'm changing the question to `blockScope wasMethodInlined` in my new branch 👍

Issue #246, opened 2026-04-13
Adds DROpalCompiler to install methods


PalumboN:
Now, we are compiling the methods inside tests. We need an API (similar to the OpalCompiler) to install the methods in the classes (and measure things).

Ducasse:
Note that with guilleP we are working on refactoring the compiler API and introducing methodInstaller.

PalumboN:
Tx Stef! Yes, we are in contact with Guille about the changes in the compiler. We will be aligned ;)

PR #255, opened 2026-07-13
Dropal compiler

### LINKED ISSUE #254: Assignment expressions fail to compile
This method fails to compile:

```st
m
	| irow |
	irow := 0.
	southToNorthMasks at: (irow := irow + 1)
```

```st
m
	| irow |
	irow := 0.
	irow := irow + 1.
	southToNorthMasks at: irow
```

But this version is ok

PalumboN:
Starting https://github.com/Alamvic/druid/issues/246 to create an API to install compiled methods in the system and run benchmarks.

My idea is to re-compile a SMark benchemark (`benchMeteor`) and compare implementations. So far, I started to compile the class `BGMeteorBoard` and solving issues related with border cases.

In summary:
- Adding `DROpalCompiler` to use as API
- Fix https://github.com/Alamvic/druid/issues/254 by splitting writes and loads at the IR
- Fix other "small" issues (special cases)

### Without inlines

we can recompile all methods and get this result:
```
[BenchmarkGameSuite new benchMeteor] bench. 
"DRUID:    10 iterations in 5 seconds 133 milliseconds. 1.948 per second"
"ORIGINAL: 35 iterations in 5 seconds 88 milliseconds. 6.879 per second"
```

### With inlines

we get these compilation errors:
```
1. hasEastOrWestIsland:
Error: Non-local return on inlined method breaks the semantics
2. northIslandsFor:row:
Error: Invalid operand: a DRPhiFunction(t52 := Ø a DRNullValue a DRNullValue) in: a DRPhiFunction(t221 := Ø t52 t221)
3. findIsland:
Error: Non-local return on inlined method breaks the semantics
4. fillMaskStartingAt:stoppingAbove:ifFoundEnough:
Error: Non-local return on inlined method breaks the semantics
5. hasSouthIsland:
#ast was sent to nil
6. islandsFor:
Error: Non-local return on inlined method breaks the semantics
7. rotationsOf:do:
Error: Phi functions are not compiled
```

and even copying those methods, the bench cannot be executed (errors at execution). So we need to continue looking here.

### DIFF
--- a/Druid-Cogit/DRCogitCanonicaliser.class.st
+++ b/Druid-Cogit/DRCogitCanonicaliser.class.st
@@ -194,6 +194,12 @@ DRCogitCanonicaliser >> visitBitOr: aDRBitOr [
 ]
 
 { #category : 'visiting' }
+DRCogitCanonicaliser >> visitBitShift: aDRBitShift [ 
+	
+	self visitLeftShift: aDRBitShift
+]
+
+{ #category : 'visiting' }
 DRCogitCanonicaliser >> visitBitXor: aDRBitXor [ 
 	
 	self sortOperandsInDoubleInstructionIfPossible: aDRBitXor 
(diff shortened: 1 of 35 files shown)


guillep:
Nice!!!!

PalumboN:
Tests are green. I merge and continue debugging on a next PR

Issue #251, opened 2026-05-04
[EPIC] Optimize the temp vectors


PalumboN:
Now, we are creating a temp vector for each scope.

Maybe it's not necessary:
- After inlinings, some scopes could finish empty (the variables are not used anymore). Check `DRCleanScope` optimization
- Also, some temp vectors could be avoided just passing the variables in copy (they just need to be read, not write).

Issue #231, opened 2025-10-31
DRCreateBlockClosure as instruction


PalumboN:
Now, the BlockClosures are DRValue (not instructions).

Would be nice to have an instruction for this with:
- Copy parameters (from the outer scope)
- Write parameters (from the outer scope)

It exists a `DRClosureCreation` instructions used for Cogit, maybe we can reuse the model.

----

And check this comment about `innerDependencies`. That should not be necessary anymore.

> We need to re-check this. Now the blocks knows the IR, so we do not need to talk with the `inlineGenerator` here. Maybe we don't need the `innerDependencies` anymore?

_Originally posted by @PalumboN in https://github.com/Alamvic/druid/pull/218#discussion_r2481326915_

PR #257, opened 2026-07-23
Fix dropal inline

### LINKED ISSUE #230: Scope ids collisions?
I think that it is possible to have `id` collisions here. Test the scopes for the cases:

- Method with two blocks (at the same level)
- Method with two inlined message sends (at the same level)

### LINKED ISSUE #256: Compiling BGMeteorBoard class
### Without inlines

we can recompile all methods and get this result:
```
[BenchmarkGameSuite new benchMeteor] bench. 
"DRUID:    10 iterations in 5 seconds 133 milliseconds. 1.948 per second"
"ORIGINAL: 35 iterations in 5 seconds 88 milliseconds. 6.879 per second"
```

### With inlines

we get these compilation errors:
```
1. hasEastOrWestIsland:
Error: Non-local return on inlined method breaks the semantics
2. northIslandsFor:row:
Error: Invalid operand: a DRPhiFunction(t52 := Ø a DRNullValue a DRNullValue) in: a DRPhiFunction(t221 := Ø t52 t221)
3. findIsland:
Error: Non-local return on inlined method breaks the semantics
4. fillMaskStartingAt:stoppingAbove:ifFoundEnough:
Error: Non-local return on inlined method breaks the semantics
5. hasSouthIsland:
#ast was sent to nil
6. islandsFor:
Error: Non-local return on inlined method breaks the semantics
7. rotationsOf:do:
Error: Phi functions are not compiled
```

and even copying those methods, the bench cannot be executed (errors at execution). So we need to continue looking here.

PalumboN:
Fix https://github.com/Alamvic/druid/issues/256
Fix https://github.com/Alamvic/druid/issues/230

Now the compilation of `BGMeteorBoard`

```st
DROpalCompiler recompileClass: BGMeteorBoard except: #().
```
finish on all (possible) cases:
```
1. hasEastOrWestIsland:
Error: Non-local return on inlined method breaks the semantics
2. northIslandsFor:row:
Error: Non-local return on inlined method breaks the semantics
3. findIsland:
Error: Non-local return on inlined method breaks the semantics
4. fillMaskStartingAt:stoppingAbove:ifFoundEnough:
Error: Non-local return on inlined method breaks the semantics
5. hasSouthIsland:
Error: Non-local return on inlined method breaks the semantics
6. islandsFor:
Error: Non-local return on inlined method breaks the semantics
```
and give these results:
```st
[BenchmarkGameSuite new benchMeteor] bench.  
"ORIGINAL:         35 iterations in 5 seconds 88 milliseconds. 6.879 per second"
"DRUID (noinline): 10 iterations in 5 seconds 133 milliseconds. 1.948 per second"
"DRUID (inline):   12 iterations in 5 seconds 111 milliseconds. 2.348 per second"
```

We should profile the causes of:
- 30% slowdown from the original Opal to DROpal without inlines
- 20% speedup from DROpal without inlines to with them

what could be related to:
- Temp vector overhead https://github.com/Alamvic/druid/issues/251
- Accessing external instance variables https://github.com/Alamvic/druid/issues/259
- JIT incompatibilities https://github.com/Alamvic/druid/issues/258

### DIFF
--- a/Druid-Opal/DRLocalVariableInstructionScheluder.class.st
+++ b/Druid-Opal/DRLocalVariableInstructionScheluder.class.st
@@ -49,6 +49,8 @@ DRLocalVariableInstructionScheluder >> shouldIgnore: aDRInstruction [
 DRLocalVariableInstructionScheluder >> storePhiOperands: aDRPhiFunction [
 
 	| variableName loadInst cfg|
+	aDRPhiFunction hasUsers ifFalse: [ ^ self ].
+	
 	variableName := aDRPhiFunction result name asLowercase asDRValue.
 	
 	"add the new created temporary variable to the scope"
@@ -64,11 +66,14 @@ DRLocalVariableInstructionScheluder >> storePhiOperands: aDRPhiFunction [
 	(aDRPhiFunction operandsWithoutMe allSatisfy: [ :op |
 		 op result isNoResult  ]) ifTrue: [ "If they are already temporaries then it is ok"
 		^ self ].
+	
+	aDRPhiFunction operands do: [ :op | op isNullValue ifTrue: [ "1halt." ^self ] ].
 
 	aDRPhiFunction operandsWithoutMe 
 		select: [ :op | op isInstruction ]
 		thenDo: [ :op |
 		| storeInst |
+		"op isPhiFunction ifTrue: [1halt. ^ self ]."
 		storeInst := DRStoreTemporaryVariable
 			             operands: {
 					             variableName.
--- a/Druid-Opal/DRLocalVariableInstructionScheluder.class.st
+++ b/Druid-Opal/DRLocalVariableInstructionScheluder.class.st
@@ -67,7 +67,7 @@ DRLocalVariableInstructionScheluder >> storePhiOperands: aDRPhiFunction [
 		 op result isNoResult  ]) ifTrue: [ "If they are already temporaries then it is ok"
 		^ self ].
 	
-	aDRPhiFunction operands do: [ :op | op isNullValue ifTrue: [ "1halt." ^self ] ].
+"	aDRPhiFunction operands do: [ :op | op isNullValue ifTrue: [ ""1halt."" ^self ] ]."
 
 	aDRPhiFunction operandsWithoutMe 
 		select: [ :op | op isInstruction ]
--- a/Druid-Opal/DRMethodIRGenerator.class.st
+++ b/Druid-Opal/DRMethodIRGenerator.class.st
@@ -27,7 +27,7 @@ DRMethodIRGenerator >> addAssignmentInstructionFrom: anAssignmentNode instructio
 		store scope: scope.
 		load scope: scope ].
 
-	^ load
+	^ store
 ]
 
 { #category : 'building' }
--- a/Druid/DRAbstractInstruction.class.st
+++ b/Druid/DRAbstractInstruction.class.st
@@ -248,7 +248,7 @@ DRAbstractInstruction >> validateDependenciesAreUsedByMyself [
 
 	self dependencies do: [ :op | 
 		(op hasUser: self) ifFalse: [ 
-			DRError signal:
+			Error signal:
 				'Dependency-user mismatch between: ' , self asString , ' and: '
 				, op asString ] ]
 ]
--- a/Druid/DRIRGenerator.class.st
+++ b/Druid/DRIRGenerator.class.st
@@ -104,6 +104,12 @@ DRIRGenerator >> addInstructionWithNoResultFrom: aNode instructionKind: instruct
 	^ self addInstruction: instruction from: aNode
 ]
 
+{ #category : 'accessing' }
+DRIRGenerator >> allTemporaryInstructions [
+
+	^ executionState allTemporaryInstructions
+]
+
 { #category : 'factory' }
 DRIRGenerator >> allocateTemporaryRegister [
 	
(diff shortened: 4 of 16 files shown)


Druid-Opal/DRLocalVariableInstructionScheluder.class.st
```diff
@@ -64,11 +66,14 @@ DRLocalVariableInstructionScheluder >> storePhiOperands: aDRPhiFunction [
 	(aDRPhiFunction operandsWithoutMe allSatisfy: [ :op |
 		 op result isNoResult  ]) ifTrue: [ "If they are already temporaries then it is ok"
 		^ self ].
+	
+"	aDRPhiFunction operands do: [ :op | op isNullValue ifTrue: [ ""1halt."" ^self ] ]."
```
guillep: Code!

Druid-Opal/DRLocalVariableInstructionScheluder.class.st
```diff
@@ -64,11 +66,14 @@ DRLocalVariableInstructionScheluder >> storePhiOperands: aDRPhiFunction [
...
 
 	aDRPhiFunction operandsWithoutMe 
 		select: [ :op | op isInstruction ]
 		thenDo: [ :op |
 		| storeInst |
+		"op isPhiFunction ifTrue: [1halt. ^ self ]."
```
guillep: Code!

Druid-Opal/DRMethodIRGenerator.class.st
```diff
@@ -27,7 +27,7 @@ DRMethodIRGenerator >> addAssignmentInstructionFrom: anAssignmentNode instructio
 		store scope: scope.
 		load scope: scope ].
 
-	^ load
+	^ store
```
guillep, line 30: Oh, looks like a nice catch

Druid-Opal/DRMethodIRGenerator.class.st
```diff
@@ -391,6 +407,12 @@ DRMethodIRGenerator >> setupCFGScope: aCodeNode [
 		node: aCodeNode
 ]
 
+{ #category : 'as yet unclassified' }
+DRMethodIRGenerator >> temporariesListfrom: aNode [ 
+	
+	^ aNode temporaryNames.
```
guillep:
```suggestion
	^ aNode temporaryNames
```

Druid/DRAbstractInstruction.class.st
```diff
@@ -248,7 +248,7 @@ DRAbstractInstruction >> validateDependenciesAreUsedByMyself [
 
 	self dependencies do: [ :op | 
 		(op hasUser: self) ifFalse: [ 
-			DRError signal:
+			Error signal:
```
guillep, line 251: Why dis?

Druid/DRInline.class.st
```diff
@@ -35,7 +35,7 @@ DRInline >> doApply: cfg [
 
 	self class enabled ifFalse: [ ^ self ].
 
-	(cfg messageSends takeFirst: 5) do: [ :messageSend | 
+	(cfg messageSends takeFirst: 3) do: [ :messageSend |
```
guillep, line 38: 3? 5? 7?

Druid/DRIRGenerator.class.st
```diff
@@ -1706,8 +1718,14 @@ DRIRGenerator >> messageSendInstructionFor: aRBMessageNode receiver: receiver ar
 	instruction inlineGenerator: inlineGenerator.
 	instruction methodNode: (method ifNotNil: [ method ast ]).
 
-	"Set the send as user to the dependecies from inner blocks"
-	instruction innerDependencies do: [ :i | i addUser: instruction ].
+	"Sends with blocks must keep the context (temporaries in the stack)
+	  That is having all outer temporaries in the dependency-user chain.
+	 If not, we don't need them, and the state is clear from previous context
+	  This simplify the inlinings, the reason why the send has a inlineGenerator
```
guillep, line 1724: Really? even in the presence of temp vectors?

guillep (dismissed):
Some small questions!

Druid-Opal/DRLocalVariableInstructionScheluder.class.st
PalumboN:
```suggestion
```

Druid-Opal/DRLocalVariableInstructionScheluder.class.st
PalumboN:
```suggestion
```

Druid/DRAbstractInstruction.class.st
PalumboN: DRError does not exist 🤷 I checked the other validations and use Error. It looks like we didn't hit this problem for a long time

Druid/DRInline.class.st
PalumboN: Image

Druid/DRIRGenerator.class.st
PalumboN: Yes, but during the compilation we need to track the def-use chain in the SSA. If not, variables used inside a block could be "dead-coded" before inlining.

The problem was that the implementation kept track a lot of trash (values that were not used), adding noise to the debugging.

Anyway, we need to revisit and re-think these decisions when optimizing temp vectors
https://github.com/Alamvic/druid/issues/231
https://github.com/Alamvic/druid/issues/251

PR #261, opened 2026-08-07
Inlining visualization

PalumboN:
Adding visualization about inlining methods. Check the video where I play with the Inline optimization:

https://github.com/user-attachments/assets/9a34dd68-aeaf-46be-ac78-cc142deaccaf

(red nodes means that there are inlining opportunities. This is a recursive method, so there are always any)


luc-raz:
Wow, this is amazing for debugging :rocket: :fire:

Druid/DRInline.class.st
```diff
@@ -33,10 +33,12 @@ DRInline class >> initialize [
...
 DRInline >> doApply: cfg [
 
+	| sends |
 	self class enabled ifFalse: [ ^ self ].
 
-	(cfg messageSends takeFirst: 3) do: [ :messageSend | 
-		self inline: messageSend ]
+	sends := cfg messageSends select: [ :send | send hasMissingInlines ].
+
+	(sends takeFirst: 3) do: [ :messageSend | self inline: messageSend ]
```
luc-raz, line 41: What the constant 3 means here

PalumboN: Hehe I'm doing experiments, that 3 is a magic number that makes the inlining strategy useful.
However, I'm changing it for a "better heuristic" 👀

PR #262, opened 2026-09-02
Compiling method with pragma

### LINKED ISSUE #243: Compiling primitive methods
This method

```st
examplePrimitiveErrorCodeModule
	"Primitive. Attempt to load a module of the given name.
	Fail if module cannot be found, or cannot be loaded,
	or failed to initialize"

	<primitive: 'primLoadModule'  module: '' error: #errorCode >
	^ self primitiveFailed
```

should be compiled as a primitive.

Check the diff:

Image

luc-raz:
Fix: #243 
Now, can compile successfully pragma with only simple primitive (eg: <primitive:1>)

For the case of pragma that defines an error variable, we need to define a temp for it. But for now, we are always using temp vector for handling all the temps (in the picture). So it breaks the execution of the method.
We need to rework on it after the optimisation of the use of the temp vector.

pragmas


Druid-Opal/DRBlockIRGenerator.class.st
```diff
@@ -18,6 +18,18 @@ DRBlockIRGenerator >> finishCodeInterpretation: lastFrame [
...
+	
+	self setupCFGScope: aDRMethod codeNode.
+	
+	self pushFrameForCode: aDRMethod receiver: aReceiver.
+	aDRMethod executeOn: self.
+	
+	^ self popFrame.
+
+]
+
```
PalumboN, line 32: This method is already implemented in the superclass `DRMethodIRGenerator`

Druid/DRMethodControlFlowGraph.class.st
```diff
@@ -45,6 +45,12 @@ DRMethodControlFlowGraph >> allInstructionsWithScope: aDRScope [
 	^ cfgInstructions , blockInstructions
 ]
 
+{ #category : 'queries' }
+DRMethodControlFlowGraph >> initialize [ 
+
+	super initialize .
+]
+
```
PalumboN: Unnecessary code

Druid-Opal-Tests/DrOpalExamples.class.st
```diff
@@ -702,6 +702,12 @@ DrOpalExamples >> simpleMethodWithBlockWritingOuterVariable: arg [
 	
 ]
 
+{ #category : 'as yet unclassified' }
+DrOpalExamples >> simpleMethodWithPragmaPrimitive [
+	<primitive:1>
+	
+]
+
```
PalumboN, line 712: Primitive 1 is the _primitiveAdd_. Since this object is not a number it is failing, and the fallback (code of the method) returns `self` (default behaviour).

So I suggest:
- To add a `self error: 'Primitive Fail'` to the code, so the test is sure that the primitive is called. Now the test should fail.
- Use a primitive that do work.  For example, `primitive: 70` is the primitiveNew, if we put the method in the class side, it should create a new instance.

Here the list of primitives:
https://github.com/pharo-project/pharo-vm/blob/8a9b398202957ff7c5c0f805ad72caf62605da6c/smalltalksrc/VMMaker/StackInterpreter.class.st#L988

Druid/DRInstructionFactory.class.st
```diff
@@ -138,6 +138,18 @@ DRInstructionFactory >> initialize [
...
+	
+	^ DRInitializePragma pragmas.
+]
+
+{ #category : 'initialization' }
+DRInstructionFactory >> initializeScope: aDRScope [
+
+	^ DRInitializeScope scope: aDRScope
+]
+
```
PalumboN: These methods must be extension methods of `Druid-Opal` package, because these classes are in that package. And sometimes we want to load Druid code without the Opal extension ;)

Druid-Opal/DRBytecodeGenerator.class.st
```diff
@@ -283,6 +283,22 @@ DRBytecodeGenerator >> visitGreaterThan: aDRGreaterThanComparison inBranch: aDRB
...
+	builder irPrimitive: aDRInitializePragma primitiveFromPragma .
+	
+	aDRInitializePragma pragmas do: [ :each | 
+		each acceptVisitor: self .
+		
+		"if the pragma is a primitive that defines an error variable, we need to store error value"
+		each isPrimitiveErrorPragma ifTrue: [
+			builder
+				storeRemoteTemp: each primitiveErrorVariableName
+				inVector: aDRInitializePragma scope tempVectorName.]
```
PalumboN: Ahh the error variable is working then? 🤔

Druid-Opal/DRInitializePragma.class.st
```diff
@@ -0,0 +1,69 @@
...
+DRInitializePragma >> scope [
+
+	^ scope
+]
+
+{ #category : 'accessing' }
+DRInitializePragma >> scope: anObject [
+
+	scope := anObject
+]
```
PalumboN: The scope of the pragma is always the method scope. The instructions already knows their CFG

`self controlFlowGraph scope`

Druid-Opal/DRBytecodeGenerator.class.st
```diff
@@ -398,6 +414,12 @@ DRBytecodeGenerator >> visitPhysicalRegister: aDRPhysicalRegister [
 	1 halt
 ]
 
+{ #category : 'API' }
+DRBytecodeGenerator >> visitPragmaNode: aPragmaNode [ 
+	
+	builder addPragma: aPragmaNode pragma.
+]
```
PalumboN, line 421: Ahh ok, this is for adding it to the metadata in the literals

Druid-Opal/DRMethodIRGenerator.class.st
```diff
@@ -90,6 +90,16 @@ DRMethodIRGenerator >> initializeIR [
...
 
+{ #category : 'initialization' }
+DRMethodIRGenerator >> initializePragma: pragmas [
+
+	| instr |
+	instr := self currentBasicBlock initializePragma.
+	instr pragmas: pragmas.
+	instr scope: self scope.
+	pragmas do: [ :each | each acceptVisitor: self].
+]
```
PalumboN, line 100: Would not be better to have one instructions per pragma? `DRPragma`s instead of `Init...`?

Druid-Opal/DRMethodIRGenerator.class.st
```diff
@@ -171,6 +181,7 @@ DRMethodIRGenerator >> interpretAssignmentNode: aRBAssignmentNode [
 DRMethodIRGenerator >> interpretCode: aDRMethod receiver: aReceiver [ 
 	
 	self setupCFGScope: aDRMethod codeNode.
+	self initializePragma: aDRMethod codeNode pragmas.
```
PalumboN, line 183: I think that the CFG should only have a InitPragma instruction if there are pragmas in the method, and not always.

Also, here we manage any pragma, not only primitives... I don't know if that is good or bad, just saying 😇

Druid-Opal/DRMethodIRGenerator.class.st
```diff
@@ -482,6 +493,19 @@ DRMethodIRGenerator >> visitInstanceVariableNode: aRBVariableNode [
...
+	
+	| var varNode|
+	"if the pragma defines an error variable, we need to declare a temp and add it to scope"
+	
+	aPragmaNode isPrimitiveErrorPragma ifFalse: [ ^ self ].
+	varNode := OCVariableNode named: aPragmaNode primitiveErrorVariableName.
+	var := PrimitiveErrorVariable node: varNode.
+	self ir scope addTempVariables: { var }
+	
+]
```
PalumboN, line 506: Ohhh our scopes have AST nodes as variables, I didn't remember that :P

PalumboN (commented):
Hi @luc-raz ,

I left some comments, answer them if you want to discuss something.

The main issue are the tests (that are not calling the primitive) and some code cleaning.

The thing to think about is the model of the pragmas in our IR. I think that an instruction per pragma is better, but I'm not sure.

PalumboN:
Ahhh and I don't understand if the error variable is working or not, because you said not but there are changes related to that.

If necessary, we can move that to another issue, and optimize the temp vectors before.

Druid-Opal/DRBlockIRGenerator.class.st
luc-raz: On the super method, the difference is I initialize the pragma

Druid-Opal-Tests/DrOpalExamples.class.st
luc-raz: Done

Druid-Opal/DRBlockIRGenerator.class.st
PalumboN: But you are not doing that in this method 🤔

Druid-Opal/DRMethodIRGenerator.class.st
PalumboN: We can open an issue for this

Issue #268, opened 2026-09-13
Generate clean and valid CFG 


PalumboN:
This code is repeated in
`DRAbstractCompilerCompiler >> generateDruidIRForAll: compiledMethods`
`DRInline >> inline: messageSend`
and here
👎 

In a good world, this should not be necessary, and the IR is generated in a valid way...

_Originally posted by @PalumboN in https://github.com/Alamvic/druid/pull/265#discussion_r3990337267_

PR #269, opened 2026-09-16
BlockClosure as real IR instruction

### LINKED ISSUE #229: DRBlockClosure ensureIR
Now we are generating the IR during object creation. We can avoid the `ensure` and have `generateIR` and `ir`

_Originally posted by @PalumboN in https://github.com/Alamvic/druid/pull/218#discussion_r2478468567_

### LINKED ISSUE #231: DRCreateBlockClosure as instruction

PalumboN:
Fix https://github.com/Alamvic/druid/issues/231
Fix https://github.com/Alamvic/druid/issues/229

Adding `DRBlockClosure` as instructions inside basic blocks (before, it only appears as message operands).

- For the method compiler, the DRBlockClosure is added to the CFG as another instruction
- It receives as operand all temporaries -> so all of them keep _used_. We can continue optimizing this in the next iterations
- Thanks that, I could remove all `innerDependecies` from the message send instructions 🧼 
- It also generates the `blockCFG` during IR generation (at the beginning, not in the end) and attached to the instruction, so it can be used by optimizations or analysis passes. I removed the `ensureIR`.
- The closures remain in the CFG after inlining. I created a `DRCleanInlinedClosures` pass to do in the end of the pipeline, but I'm not sure about it


Druid/DRBlockClosure.class.st
```diff
@@ -85,13 +86,13 @@ DRBlockClosure >> hash [
...
+1halt.
 	^ irGenerator
 ]
 
 { #category : 'accessing' }
 DRBlockClosure >> irGenerator: anIRGenerator [
-
+1halt.
 	irGenerator := anIRGenerator
 ]
```
PalumboN: Check this and remove `irGenerator` variables if not used

Druid-Opal/DRBlockIRGenerator.class.st
```diff
@@ -18,6 +18,26 @@ DRBlockIRGenerator >> finishCodeInterpretation: lastFrame [
...
+	self numberOfArguments: numArgs.
+	self interpretCode: aDRBlockClosure receiver: aDRBlockClosure.
+
+	cfg := self ir.
+	cfg numArgs: numArgs.
+
+	"Clean garbage to generate a valid CFG"
+	DRDeadBlockElimination new applyTo: cfg.
+	DRDeadCodeElimination new applyTo: cfg.
+	cfg validate.
```
luc-raz, line 36: we do these 3 instructions in differents places
Would it be nice to refactor it in a method?

guillep: @PalumboN tells me that it is this issue: https://github.com/Alamvic/druid/issues/268 :)

guillep:
Thanks @luc-raz for the review!

